PR #3, opened 2016-12-20
Making GeoPySpark Self Contained

jbouffard:
This PR updates the GeoPySpark project so that it is self contained such that all required dependencies and files can be found within it. To make this project more "demo" like, a Makefile was also added to make testing running the project easier.


jbouffard:
@lossyrob This is ready for review.

PR #4, opened 2016-12-21
SpatialKey Now Needs to be Installed

jbouffard:
This PR corrects the issue of python not being able to find the `SpatialKey` module. This is done by having the user install it locally before `spark-submit` runs. In addition, a few new commands have been added to the `Makefile`, and there has been a slight restructuring of the project layout.


jbouffard:
@lossyrob This is ready for review.

Makefile
```diff
@@ -1,8 +1,17 @@
+install-spatialkey:
+	cd spatial_key && \
+		unzip python_spatial_key.zip && \
```
lossyrob: Let's not use a zip file, and instead have the SpatialKey type inside of `geopyspark`. So users would have to run `python setup.py install` in the root of the repo, and those types would be exposed.

jbouffard:
@lossyrob Made the recommended changes. Should be good to merge now.

PR #5, opened 2016-12-22
More Avro Serilaization

jbouffard:
This PR adds more Avro serialization which allows for values to be passed from scala to python and back.


jbouffard:
@lossyrob Ready for review

PR #28, opened 2017-01-30
[WIP] Refactoring the AvroSerializer

jbouffard:
This PR seeks to add the ability to keep track of all use defined classes and their respective encoding/decoding methods.


jbouffard:

geopyspark/avroserializer.py
```diff
@@ -10,13 +9,10 @@
 class AvroSerializer(FramedSerializer):
     def __init__(self,
             schema_json,
-            geotrellis_decoder=GeoTrellisDecoder(),
-            geotrellis_encoder=GeoTrellisEncoder()):
+            avroregistry=AvroRegistry()):
```
echeipesh: This is a gotcha. The registry is going to be created once during method definition and re-used in all constructors. Any state change to this object will be available to all future instance of `AvroSerializer`. The pattern around it seems to be to set the default value to `None` and check for that in the body. (ref: http://docs.python-guide.org/en/latest/writing/gotchas/)

But `AvroRegistry` should have a `DEFAULT` fields to avoid multiple re-instantiations which may be expensive in this case.

geopyspark/avroserializer.py
```diff
@@ -10,13 +9,10 @@
 class AvroSerializer(FramedSerializer):
     def __init__(self,
             schema_json,
```
echeipesh: This and the related usages of schema are no longer needed now that we have `avroregistry` to fetch decoders and encoders?

geopyspark/geotrellis/geotiff_rdd.py
```diff
@@ -2,117 +2,106 @@
...
+from geopyspark.avroregistry import AvroRegistry
 
 
 class GeoTiffRDD:
-    def set_custom_class(self, custom_class):
-        self.custom_class = custom_class
-
-    def _decode_java_rdd(self, java_rdd, schema):
-        if self.custom_class is None:
+    def _decode_java_rdd(self, java_rdd, schema, avroregistry):
```
echeipesh: Again schema should no longer be needed here?

geopyspark/avroserializer.py
jbouffard: We still need that schema. It needs to be decoded by Avro so that we know what the types are in the `RDD`. https://github.com/jbouffard/geopyspark/blob/refactor-3-the-reckoning/geopyspark/avroserializer.py#L31

geopyspark/geotrellis/geotiff_rdd.py
jbouffard: See answer to the other comment.

geopyspark/avroserializer.py
jbouffard: Just fixed the issue.

geopyspark/geotrellis/geotiff_rdd.py
echeipesh: Thank you, I missed that usage, but it makes total sense.

PR #31, opened 2017-02-03
Adding Travis CI to GeoPySpark

### LINKED ISSUE #30: Dangerous manipulation of path strings
In python, you'd typically want to manipulate paths with the [`os` package](https://docs.python.org/3/library/os.html). Working directly with strings is brittle and semantically opaque - we should make this switch when/where possible

jbouffard:
This PR adds Travis CI to GeoPySpark. In order to achieve this, the way in which tests are run and implemented had to be changed. There are still further tests that can be implemented by Travis (PEP8, Pylint, et,), but this PR will ensure that any future PR will work and be compatible with Master.

In addition to adding Travis, this PR also resolves #30


jbouffard:

PR #32, opened 2017-02-06
Catalog Refactor

jbouffard:
This PR refactors both the `catalog` API and its scala backend to make them more pythonic.


jbouffard:
@echeipesh This is ready for review. I decided not to include any of the layer tests as I think that is another PR in it of itself.

PR #36, opened 2017-02-16
GeoPyContext

jbouffard:
This PR adds `GeoPyContext` class that is a wrapper for `SparkContext`. This class will provide various `JavaClass`es to be found in one place.

PR #37, opened 2017-02-17
GeoPyRDD

jbouffard:
This PR is based on another PR that is still in review, #36 . A new classes has been added called, `GeoPyRDD`. This is a subclass of `PySpark`'s `RDD` class, and simply contains additional information, such as the `RDD`'s `schema`, so that it is easier to access for late use.


jbouffard:
@echeipesh Once #36 is merged in, this will be ready for review.

PR #38, opened 2017-02-17
Update build and Dockerfile

echeipesh:
These build changes are meant to better support building a Jupyter notebook docker file that can be used as an interface.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,5 @@ pytest>=3.0.6
 numpy>=1.8
 avro-python3>=1.8
 shapely>=1.6b3
-rasterio>=0.36.0
 py4j>=0.10.4
 setuptools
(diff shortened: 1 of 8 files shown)


requirements.txt
```diff
@@ -2,6 +2,5 @@ pytest>=3.0.6
 numpy>=1.8
 avro-python3>=1.8
 shapely>=1.6b3
-rasterio>=0.36.0
```
jbouffard: It appears that taking Rasterio out has caused Travis to fail: https://travis-ci.org/locationtech-labs/geopyspark/jobs/202749661#L4707

Are you installing it else where? Rasterio is not needed for any of the actual code to function, however, it's required for the tests.

jbouffard:
:+1:

PR #41, opened 2017-02-21
Test Fixes

jbouffard:
This PR fixes the various unit tests by having only one `SparkContext` for all of the tests. By doing this, all tests can now be ran as one batch instead of some of them being run separately.


jbouffard:

PR #45, opened 2017-02-24
Jar Distribution

jbouffard:
This PR is based on another PR that is still in review, #41 . The main feature added is that the `jar` file that is needed to drive GeoPySpark will now be distributed in the wheel. In addition, PySpark environment variables have now been set in a new way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ export IMG := jupyter-geopyspark
 
 ASSEMBLY := geopyspark-backend/geotrellis/target/scala-2.11/geotrellis-backend-assembly-0.1.0.jar
 WHEEL := dist/geopyspark-0.1.0-py3-none-any.whl
-JAR-PATH := geopyspark/jar/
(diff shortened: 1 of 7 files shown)


jbouffard:
@echeipesh This will be ready for review once #41 is merged in.

Makefile
```diff
@@ -4,20 +4,24 @@ export IMG := jupyter-geopyspark
...
 	${PYTHON} setup.py bdist_wheel
 
-wheel: ${WHEEL}
+wheel: ${ASSEMBLY} ${JAR-PATH} ${WHEEL}
 
-pyspark: ${ASSEMBLY}
+pyspark:
 	pyspark --jars ${ASSEMBLY}
 
 docker-build: ${WHEEL}
```
jpolchlo: This line should read `docker-build: wheel`.

echeipesh: Does that make a difference? `wheel` is a phantom target as well.

PR #46, opened 2017-03-02
Spark Environment Variable Fix

jbouffard:
This PR fixes an issue that was never addressed in #45 . Not having `pyspark` imported before importing a `GeoPySpark` class or function that relies on `pyspark` would cause an error. Now, `pyspark` will be added the `sys.path` during the setup of the other environment variables. Thus, prevent that error from occurring.


jbouffard:

PR #47, opened 2017-03-06
Multiband Fix

jbouffard:
This PR changes how `Multiband`s are returned in python. Instead of getting back a `[TileArray, TileArray, .., TileArray]`, the use now gets back a single `TileArray` that has a shape `(n, cols, rows)`. Where `n` is the number of bands in that array. The reason for this change was to match what Rasterio returns when it reads in a `Multiband` file.


jbouffard:

PR #49, opened 2017-03-06
Schema Producer

jbouffard:
This PR allows the user to create an `AvroSchema` from the python side given `K` and `V` within the `RDD`.


jbouffard:
Closed this issue since it was not meant to be against `master`.

PR #51, opened 2017-03-08
Serialization Refactor - Encoding and Decoding

jbouffard:
This PR is based on another, open Pr, #50 . Encoding and Decoding have been refactored. The biggest changes is that now there is only one method to do a task as opposed to four. For example, instead of having: `query_spatial_singleband`, `query_spatial_multiband`, `query_spacetime_singleband`, and `query_spacetime_multiband`, it is now just: `query`. The changes to the internal code to required to make this refactor work have also allowed for `RDD`s made in python to be sent to scala.

The refactor is still ongoing, and this PR should not be merged unless the one that comes next is merged right after.


jbouffard:

PR #52, opened 2017-03-09
Serialization Refactor - More Decoder and Encoder Refactoring

jbouffard:
This PR is based on another, open Pr, #51 . Decoding and Encoding are now further refactored. The only decoding and encoding methods available now are for `Tile`s, `MultibandTile`s, and `Tuple2`s. Everything else is read in and out as is. Additional support has been added to allow for the use of custom encoders and decoders.

While the refactor is still continuing, this Pr is stable, and can be merged in without issue.

This PR resolves #43

Issue #48, opened 2017-03-06
Return TileLayerMetadata As a dict When Using the Catalog


jbouffard:
Currently, we return `TileLayerMetadata` as a instance of the `Metadata` class when reading layers using one of the catalogs [link](https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/catalog.py#L9). However, this should be changed so that the all the information is returned in a `dict` like we did when collecting the metadata [link](https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/tile_layer_methods.py#L83).

jbouffard:
Resolved in #53

PR #53, opened 2017-03-09
Serialization Refactor - Catalog Metadata

jbouffard:
This PR is based on another, open Pr, #52 . When reading or querying data with one of the catalogs, the return is now `(RDD, schema, metadata)` as opposed to `(RDD, Metadata)`. This will make working with the various catalogs easier.

While the refactor is still continuing, this Pr is stable, and can be merged in without issue.

This Pr resoles #48


jbouffard:

Issue #59, opened 2017-03-14
Create Constants for the KeyType of RDDs


jbouffard:
Currently, we have the user manually input either `"spatial"` or `"spacetime"` to designate the type of the key within the `RDD` that is being performed on. Instead, we should use constants to make things easier on the user.

jbouffard:
Resolved by #61

PR #61, opened 2017-03-14
GeoTrellis Constants

jbouffard:
This Pr adds GeoTrellis constants to GeoPySpark. This will make working with methods that require `rdd_type` easier as it restricts it to just two constants instead of strings.

This Pr fixes issue #59


jbouffard:

echeipesh:
The file name of `geopyspark.geotrellis.geotrellis_constants.py` makes the import a little redundant. Unless there is an argument otherwise it should just be ` geopyspark.geotrellis.constants.py` like in the mock up API.

jbouffard:
@echeipesh Resolved file name issue.

PR #65, opened 2017-03-15
Tile Encoder/Decoder Update

### LINKED ISSUE #60: Have All Tiles Be Decoded/Encoded as a MultibandTile
As of now, there are two different ways to read `Tile`s in and out depending on its bound count. However, reading a `Tile` as a `MultibandTile` still produces the correct result, and the user interact with the underlying reading methods. Therefore, it'd be best to collapse `Tile`s and `MultibandTile`s into one decode and encode method. This will simplify the API, and make the underlying code easier to read and work with.

jbouffard:
This Pr merges `Tile` and `MultibandTile` so that they are both treated the same in GeoPySpark. Now, `Tile`s will be read in and sent to scala as `MultibandTile`s regardless of the number of bands they have. This greatly simplifies the code, and makes it easier to work with.

This Pr fixes #60

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
@@ -8,6 +8,7 @@ import geotrellis.spark.io.cassandra._
 import geotrellis.spark.io.file._
 import geotrellis.spark.io.hadoop._
 import geotrellis.spark.io.hbase._
+import geotrellis.spark.io.json._
 import geotrellis.spark.io.s3._
 import geotrellis.vector._
 import geotrellis.vector.io.wkt.WKT
@@ -15,6 +16,8 @@ import geotrellis.vector.io.wkt.WKT
 import org.apache.spark._
 import org.apache.spark.api.java.JavaRDD
 import org.apache.spark.rdd.RDD
+
+import spray.json._
 
 import java.time.ZonedDateTime
 import java.util.ArrayList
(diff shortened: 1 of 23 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
```diff
@@ -121,33 +110,20 @@ abstract class FilteringLayerReaderWrapper()
...
-        PythonTranslator.toPython(query.result)
-      }
-      case("SpatialKey", "MultibandTile") => {
+    keyType match {
+      case "SpatialKey" => {
         val spatialQuery = getSpatialQuery(queryGeometryString)
 
         val layer = layerReader.query[SpatialKey, MultibandTile, TileLayerMetadata[SpatialKey]](id)
         val query = spatialQuery match {
+          case Some(point: Point) => layer.where(Contains(point))
```
jbouffard: @echeipesh Will we still be able to query `Point`s within a `MultibandTile`? Because the original code only matched against `Point`s when `V` was a `Tile`.

jbouffard:

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
echeipesh: Yes, this "just work" for `MultibandTile` from geotrellis end

jbouffard: Okay. Then there's nothing else that I had any questions on.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
```diff
@@ -57,26 +55,17 @@ abstract class FilteringLayerReaderWrapper()
...
 
-    (keyType, valueType) match {
-      case ("SpatialKey", "Tile") => {
-        val results = layerReader.read[SpatialKey, Tile, TileLayerMetadata[SpatialKey]](id)
-        PythonTranslator.toPython(results)
-      }
-      case ("SpatialKey", "MultibandTile") => {
+    keyType match {
+      case "SpatialKey" => {
         val results = layerReader.read[SpatialKey, MultibandTile, TileLayerMetadata[SpatialKey]](id)
```
echeipesh: We still need to decide if we're doing `Tile` or `MultibandTile` query for geotrellis because the Avro schema used in storage is different. So what has to happen here is `val header = attributeStore. readHeader[SpatialKey]` and inspect `valueClass` String. If the layer stores `Tile` we must query for tile and then wrap them in `ArrayMultibandTile` to return to python. Since this has to happen at every read/query it should be made generic enough to be a function.

geopyspark/avroregistry.py
```diff
@@ -106,19 +100,15 @@ def tile_encoder(obj):
         return datum
 
     @classmethod
-    def multiband_encoder(cls, obj):
+    def tile_encoder(cls, obj):
+        band_count = obj['data'].shape[0]
 
         def create_dict(index):
```
echeipesh: Does this need reshape `obj` if its only a `2d` array ?

jbouffard: Yeah, as of right now, you need to use 3d arrays only.

echeipesh: Ok, we should log this as an issue if we're not going to address this right now.

jbouffard:
@echeipesh I fixed the issue with `read` and `query` for the catalog and made it so that you can now decode/encode 2D numpy arrays.

PR #67, opened 2017-03-15
Geotiff Rdd Refactor

### LINKED ISSUE #63: Refactor geotiff_rdd API
`geotiff_rdd` contains two classes, `HadoopGeoTiffRDD` and `S3GeoTiffRDD`, respectively. It would be beneficial to combine these into one method, `read_geotiff_rdd`. The backend would be determined by the `uri` given. This would simplify things greatly, and allow for easier use of GeoPySpark.

Usage of this API can be seen in #62

jbouffard:
This is Pr is based on another, open Pr #65 . `HadoopGeoTiffRDD` and `S3GeoTiffRDD` have been removed in favor of just a single method, `geotiff_rdd`. By taking a `uri`, this function will be able to determine which backend to read from.

This Pr fixes #63

### DIFF
--- a/geopyspark/geotrellis/geotiff_rdd.py
+++ b/geopyspark/geotrellis/geotiff_rdd.py
@@ -1,37 +1,43 @@
 from geopyspark.geopycontext import GeoPyContext
 
 
-class HadoopGeoTiffRDD(object):
-    def __init__(self, geopysc):
-        self.geopysc = geopysc
-        self._hadoop_wrapper = self.geopysc.hadoop_geotiff_rdd
+def geotiff_rdd(geopysc,
+                rdd_type,
+                uri,
+                options=None,
+                **kwargs):
 
-    def get_rdd(self, key_type, path, options=None):
-        key = self.geopysc.map_key_input(key_type, False)
+    key = geopysc.map_key_input(rdd_type, False)
 
-        if options is None:
-            result = self._hadoop_wrapper.getRDD(key, path, self.geopysc.sc)
+    if kwargs and not options:
+        options = kwargs
+
+    if 's3' in uri:
+        key_and_bucket_uri = uri.split("s3://")[1]
+        (bucket, prefix) = key_and_bucket_uri.split("/", 1)
+
+        if not options:
+            result = geopysc.s3_geotiff_rdd.getRDD(key,
+                                                   bucket,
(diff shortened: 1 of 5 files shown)


jbouffard:
@echeipesh This will be ready for review once #65 is merged.

geopyspark/geotrellis/geotiff_rdd.py
```diff
@@ -1,37 +1,86 @@
...
-            result = self._hadoop_wrapper.getRDD(key, path, options, self.geopysc.sc)
+    """Creates a RDD from GeoTiffs that are located on the local file system, HDFS, or S3.
+
+    Args:
+        geopysc (GeoPyContext): The GeoPyContext being used this session.
+        rdd_type (str): What the spatial type of the geotiffs are. This is
+            represented by the constants: SPATIAL and SPACETIME. Note: All of the
+            GeoTiffs must have the same saptial type.
+        uri (str): The path to a given file/directory.
+        options (dict, optional): A dictionary of different options that are used
```
echeipesh: Available options should defiantly be included in the doc string, otherwise there is no way for people to know

jbouffard: @echeipesh Just updated the docs.

geopyspark/geotrellis/geotiff_rdd.py
```diff
@@ -1,37 +1,114 @@
...
+    """
 
-        if options is None:
-            result = self._s3_wrapper.getRDD(key, bucket, prefix, self.geopysc.sc)
+    key = geopysc.map_key_input(rdd_type, False)
+
+    if kwargs and not options:
+        options = kwargs
+
+    if 's3' in uri:
```
echeipesh: This does not seem sufficiently robust, S3 anywhere in the Uri will be sufficient?

echeipesh: geotrellis uses an Rd for this, should at least lift that if there isn't standard uri object for python

jbouffard: @echeipesh I just updated it, but there's another alternative that's probably closer to what you're working for. Though, it's a bit more work.

It's look like: `if urlparse(uri).scheme == 's3:`

PR #68, opened 2017-03-17
Tile layer Refactor

### LINKED ISSUE #64: Refacoter TileLayerMethods API
`TileLayerMethods` currently exists as a class containing various methods. It's been decided that these methods should be taken out of the class and be accessed individually. We believe this will help improve the API of GeoPySpark. An example of this can be seen in #62

jbouffard:
This is Pr is based on another, open Pr #67 . `TileLayerMethods` is now removed and its former methods are now independent functions in `tile_layer`. Two additional functions have been added, `collect_pyramid_metadata` and `pyramid`. This will make it possible to create pyramided ingests with `RDD`s from any source.

This Pr fixes #64


jbouffard:

geopyspark/geopycontext.py
```diff
@@ -61,6 +61,10 @@ def tile_layer_methods(self):
     def tile_layer_merge(self):
         return self._jvm.geopyspark.geotrellis.spark.merge.MergeMethodsWrapper
 
+    @property
+    def pyramid_builder(self):
```
echeipesh: Why is `pyramid_builder` special enough to hang on the context?

jbouffard: I did that because that's where all of the other Scala objects are, and also because `GeoPyContext` is essentially the bridge between scala and python.

PR #72, opened 2017-03-18
Yet Another Catalog Refactor

### LINKED ISSUE #69: Refactor Catalog API
There are currently different `Catalog` classes for each type of backend. We should remove these classes and instead just keep the three functions: `read`, `query`, and `write`. This can be accomplished by using a single `uri` input format across all backends. An example of this can be seen in #63 .

By doing this, we will greatly simplify `catalog` and make it easier for our users to use.

jbouffard:
This Pr is based off another, open Pr #68 . This Pr removed the various classes from `catalog`, leaving just the `read`, `query`, and `write` functions. This was done by requiring all paths to be in a `uri` format. By having just these three methods, the code has become much more simplified and is easier to use, read, and understand. In addition to the refactor, docs have been added to `catalog`.

This Pr fixes #69 .


jbouffard:
@echeipesh This is ready for review once #68 is merged.

PR #74, opened 2017-03-20
Value reader Refactor

### LINKED ISSUE #70: Refactor Value Reader API
Like, `Catalog` `ValueReader` has a different class for each backend. It'd be best just to remove these classes and retain the single `value_reader` function. This can be achieved in the same way proposed in #69 . An example of this can be seen in #63 .

By doing this, we will greatly simplify `catalog` and make it easier for our users to use.

jbouffard:
This Pr is based off another, open Pr #72  . This Pr removed the `value_reader` module and its classes, and moved the `read_value` method over to `catalog`. Thus, now all functions that deal with layer reading/writing can be now be found in one place. This should make things easier to use and and increase performance.

This Pr fixes #70


jbouffard:

PR #75, opened 2017-03-21
Reproject

jbouffard:
This Pr is based on another, open Pr #74 . Two new reprojection functions have been added to `tile_layer` that can reproject either a `RDD` or a layer called, `reproject` and `reproject_pyramid`, respectively. This will make it possible to pyramid more types of tiles since pyramiding is only possible for certain types.


jbouffard:
@echeipesh This is ready for review once #74 is merged.

PR #76, opened 2017-03-22
Avoid calls to _to_java_object_rdd during RDD encoding

jbouffard:
This Pr is based on another, open Pr #75 . A partial fix to the performance issues with GeoPySpark. Now functions that take a `RDD` that came from Scala and then send it back will be much faster. Depending on the `RDD` and function, there was between a 50%-80% increase in speed. Making times much more manageable.

However, `RDD`s that originated in Python still have performance issues. These have been partially solved in this Pr, but not to the same extent as the scala `RDD`s. Thus, it is still recommended to use GeoPySpark to read in GeoTiffs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,6 @@ notebooks
 
 # GeoPySpark-backend jar
 geopyspark/jars/*.jar
+
+# Unit test performance results
+prof/
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/spark/TileLayerMetadataCollector.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/spark/TileLayerMetadataCollector.scala
@@ -17,6 +17,7 @@ import spray.json._
 
 import org.apache.spark._
 import org.apache.spark.rdd.RDD
+import org.apache.spark.api.java.JavaRDD
 
 import scala.collection.JavaConverters._
 import collection.JavaConversions._
@@ -31,7 +32,7 @@ object TileLayerMetadataCollector {
   T <: CellGrid: AvroRecordCodec: ClassTag,
   K2: Boundable: SpatialComponent: JsonFormat
   ](
-    returnedRdd: RDD[Array[Byte]],
+    returnedRdd: JavaRDD[Array[Byte]],
     schemaJson: String,
     pythonExtent: java.util.Map[String, Double],
     pythonTileLayout: java.util.Map[String, Int],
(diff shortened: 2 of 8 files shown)


jbouffard:
@echeipesh This is ready for review once #75 is merged.

echeipesh:
I don't really have a problem with centering API about using `JavaRDD` vs `RDD` but I'm confused why that would make any difference at all, it appears to be a dumb container: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/java/JavaRDD.scala#L30

Do you have any sense why that might be?

jbouffard:
@echeipesh There is a difference between the two.

The way we used to send a `PythonRDD` over to scala was: serialize the contents of the `PythonRDD` -> converting the serialized `PythonRDD` to a `JavaRDD` via `python_rdd._to_java_object_rdd()` -> and then convert the `JavaRDD` to a `ScalaRDD` through `java_rdd.rdd()`.

[_to_java_object_rdd()](https://github.com/apache/spark/blob/master/python/pyspark/rdd.py#L2258) takes the serialized `PythonRDD` and then serializes it again with `AutoBatchedSerializer(PickleSerializer())`. Once that's done, the actual conversion happens with [pythonToJava](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/api/python/SerDeUtil.scala#L147). Which just `unpickle`s the objects within the `PythonRDD` and then converts it to a `JavaRDD`. This is then just unwrapped via `rdd()`.

The new way involves using `_jrdd` on the `PythonRDD` to convert it to a `JavaRDD`.

When brining the `RDD` from scala, the `_jrdd` is just the wrapped `JavaRDD` that was brought over [link](https://github.com/apache/spark/blob/master/python/pyspark/rdd.py#L181). Once any method is performed on the `PythonRDD`, a new underlying `_jrdd` is created [link](https://github.com/apache/spark/blob/master/python/pyspark/rdd.py#L2429) This is done by sending the function and various parameters to scala to create a `PythonRDD` and then converting it to a `JavaRDD`.

I'm starting to wonder if these two methods produce different `JavaRDD`s of the same data, though. I'm going to revert back to the old method to see if there's a performance difference when the scale of data increases.

echeipesh:
So just to be extra clear clear this is what I think is going on:

Our scala side of the wrapper is only able to read `JavaRDD`s which contain bytes that are encoded with Avro. 

The previous calls to [`_to_java_object_rdd`](https://github.com/apache/spark/blob/master/python/pyspark/rdd.py#L2258-L2265) would trigger an extraneous conversion from `JavaRDD` of whatever bytes to `JavaRDD` of pickled bytes before they were once again converted into `JavaRDD` of Avro bytes and handed over to the wrappers. 

This PR avoids the intermediate `_reserialiation` caused by `_to_java_object_rdd`. The switch to using `JavaRDD` in wrappers actually makes the python code a little less clearer by avoiding avoiding the second nesting so I think that's fine.

PR #77, opened 2017-03-23
Jupyter Container On EMR

jamesmcclain:
First
```
sudo yum install -y docker
sudo service docker start
<log out, log back in>
sudo usermod -a -G docker hadoop
sudo yum install python34
sudo easy_install-3.4 pip
sudo /usr/local/bin/pip3 install --upgrade pip  # only if building
sudo /usr/local/bin/pip3 install wheel  # only if building
```
then (to test on EMR)
```
docker run -it --rm --net=host -v /etc/hadoop/conf:/yarn:ro quay.io/geodocker/jupyter:3 bash
```
then
```
YARN_CONF_DIR=/yarn $SPARK_HOME/bin/spark-submit --class org.apache.spark.examples.BroadcastTest --master yarn --deploy-mode client $SPARK_HOME/examples/jars/spark-examples_2.11-2.1.0.jar
```

To run Jupyter:
```
docker run -it --rm -p 8000:8000 --net=host -v /etc/hadoop/conf:/yarn:ro quay.io/geodocker/jupyter:3
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setup(
         'shapely>=1.6b3'
     ],
     include_package_data=True,
-    data_files=[('jars', ['geopyspark/jars/geotrellis-backend-assembly-0.1.0.jar'])],
+    # data_files=[('jars', ['geopyspark/jars/geotrellis-backend-assembly-0.1.0.jar'])],
     packages=['geopyspark', 'geopyspark.geotrellis', 'geopyspark.tests', 'geopyspark.jars'],
     scripts=[],
     classifiers=[
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+build/
+dist/
+examples/
+geopyspark/
+geopyspark-backend/
+geopyspark.egg-info/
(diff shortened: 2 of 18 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
         'shapely>=1.6b3'
     ],
     include_package_data=True,
-    data_files=[('jars', ['geopyspark/jars/geotrellis-backend-assembly-0.1.0.jar'])],
+    # data_files=[('jars', ['geopyspark/jars/geotrellis-backend-assembly-0.1.0.jar'])],
```
echeipesh: @jbouffard Is this the thing that included the jar in the wheel so it can be called in the vanilla case of "import geopyspark" in python process ?

jbouffard: Yeah, but I don't think that's needed since `jars` is a module within `GeoPySpark` and it will just install it by default.

echeipesh: Ok, interesting. I think that works out. I just don't want to lose the ability to run GeoPySpark easily in local mode when getting the EMR options.

jamesmcclain:
I have made the changes that I wanted to.

PR #81, opened 2017-03-29
Prevent Stale Test Data

jamesmcclain:


geopyspark/tests/python_test_utils.py
```diff
@@ -1,6 +1,6 @@
 import os
+import shutil
```
jbouffard: Sorry for nitpicking, but could you please switch `import shutil` with `from os import path` so that the former is on line 3 and the latter on line 2? The python coding standard wants `imports` of the same module to be grouped together.

jamesmcclain: Okay, will-do.

jamesmcclain:
Are there are any other similar issues (of the type fixed by this patch) that you are aware of?

jbouffard:
I think I got them all when I cleaned up the codebase awhile ago, but I might have missed a few. I'm not sure what plugins you have for python, but I have one that points out where there are style issues in my code.

jamesmcclain:
I was referring to the issue of the stale zip file (the issue fixed by this PR).  Are there any similar issues that you are aware of?

jbouffard:
Oh, not to my knowledge. I think that's the first time an issue like that has arose .

PR #84, opened 2017-03-30
Reproject Methods Refactor and Expansion

### LINKED ISSUE #82: Resolve Reprojected Issue
It appears that both `reproject_pyramid_layout` and `reproject_pyramid_zoom` are not actually reprojecting the tiles within the `RDD` correctly. This should be fixed before the initial release of GeoPySpark.

jbouffard:
This Pr seeks to expand, fix, and refactor the reprojections of GeoPySpark. Two reproject methods have been added, `reproject` and `reproject_with_floating_layout`. In addition, fixes have been made to the `reproject_with_tile_layout` method so that it now works correctly.

This Pr fixes #82

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/spark/reproject/ReprojectWrapper.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/spark/reproject/ReprojectWrapper.scala
@@ -23,21 +23,17 @@ import org.apache.spark.api.java.JavaRDD
 import scala.reflect.ClassTag
 
 object ReprojectWrapper {
-  private def reprojectRDD[
-    K: SpatialComponent: ClassTag: Boundable: AvroRecordCodec: JsonFormat
-  ](
+  private def reprojectRDD(
+    keyType: String,
     returnedRDD: JavaRDD[Array[Byte]],
     schema: String,
-    metadata: TileLayerMetadata[K],
+    returnedMetadata: String,
     destCRS: String,
     layout: Either[LayoutScheme, LayoutDefinition],
     matchLayerExtent: Boolean,
     resampleMethod: String
   ): (Int, (JavaRDD[Array[Byte]], String), String) = {
-    val rdd: RDD[(K, MultibandTile)] =
-      PythonTranslator.fromPython[(K, MultibandTile)](returnedRDD, Some(schema))
-
-    val contextRDD = ContextRDD(rdd, metadata)
+    val metadataAST = returnedMetadata.parseJson
     val crs = CRS.fromName(destCRS)
 
     val options = {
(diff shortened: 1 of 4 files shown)


jbouffard:

jamesmcclain:
Is all of the expanded functionality covered by tests?

jbouffard:
For the most part. It's a little difficult to test fully in python; but the tests should be okay.

jamesmcclain:
The reprojection tests are not passing for me locally:
```
geopyspark/tests/reproject_test.py FFFF
```

```
E                   py4j.protocol.Py4JError: An error occurred while calling z:geopyspark.geotrellis.spark.reproject.ReprojectWrapper.reproject. Trace:
E                   py4j.Py4JException: Method reproject([class java.lang.String, class org.apache.spark.api.java.JavaRDD, class java.lang.String, class java.lang.String, class java.lang.String, class java.lang.Integer, class java.lang.Double, class java.lang.String, class java.lang.Boolean]) does not exist
E                       at py4j.reflection.ReflectionEngine.getMethod(ReflectionEngine.java:318)
E                       at py4j.reflection.ReflectionEngine.getMethod(ReflectionEngine.java:339)
E                       at py4j.Gateway.invoke(Gateway.java:274)
E                       at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:128)
E                       at py4j.commands.CallCommand.execute(CallCommand.java:79)
E                       at py4j.GatewayConnection.run(GatewayConnection.java:211)
E                       at java.lang.Thread.run(Thread.java:745)

/usr/local/lib/python3.4/dist-packages/py4j/protocol.py:323: Py4JError
```

They obviously passed on Travis, but some investigation may be in order.

jbouffard:
@jamesmcclain You'll have to re-assemble the backend and then move the resulting jar to the `jars` directory.

jamesmcclain:
I ran `make install` but that did not do it, I had to do it manually.  Perhaps the makefile needs some attention.

jbouffard:
I agree. I've been meaning to add a `make` command to do that for some time. Would like me to make a Pr for it? Or would you rather do it?

jamesmcclain:
> I agree. I've been meaning to add a make command to do that for some time. Would like me to make a Pr for it? Or would you rather do it?

Either this PR or another is fine for me ... I just want to make sure that we remember to do it.

jamesmcclain:
Besides points discussed above, this looks okay to me.  Please wait for @echeipesh , though.

jbouffard:
@jamesmcclain Okay, will do. I just made it so that `make install` will now always assemble and move the jar before installation.

jbouffard:

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/spark/reproject/ReprojectWrapper.scala
```diff
@@ -48,49 +62,105 @@ object ReprojectWrapper {
...
       case "SpatialKey" => {
         val metadataAST = returnedMetadata.parseJson
         val metadata = metadataAST.convertTo[TileLayerMetadata[SpatialKey]]
-        val layout = Right(metadata.layout)
+        val layout = Right(LayoutDefinition(layoutExtent.toExtent, tileLayout.toTileLayout))
 
-        reprojectRDD[SpatialKey](returnedRDD, schema, metadata, destCRS, layout, matchLayerExtent)
+        reprojectRDD[SpatialKey](returnedRDD, schema, metadata, destCRS, layout, matchLayerExtent, returnedResampleMethod)
       }
       case "SpaceTimeKey" => {
```
echeipesh: This check is done all over the file and it looks like it can just happen in `private def reprojectRDD`

geopyspark/geotrellis/tile_layer.py
```diff
@@ -375,22 +406,63 @@ def reproject(geopysc,
...
 
-def reproject_pyramid(geopysc,
-                      rdd_type,
-                      keyed_rdd,
-                      tile_layer_metadata,
-                      dest_crs,
-                      tile_size,
-                      resolution_threshold=0.1,
-                      match_layer_extent=False):
+def reproject_with_floating_layout(geopysc,
```
echeipesh: It would be good to avoid all of these renames: `reproject_with_floating_layout`, `reproject_with_tile_layout`, `reproject_with_zoomed_layout` and have a single `reproject` method. I'm not sure I track why they are required.

jbouffard: That was mostly done to make it easier for people to understand the differences between the three functions. I can probably collapse those into one, though.

echeipesh:
From python the API should look something like this

```py
# perform per-tile reproject based on tile ProjectedExtent
.reproject("EPSG:3857")

# All the options from https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/reproject/Reproject.scala#L35 are optional or picked up from kwargs
# This can be performed on anything that has ProjectedExtent

# perform buffered reproject, this requires that keys are either SpatialKey or SpatioTemporal Keys
.reproject_to_layout("EPSG:3857",  layoutSchemeOrLayoutDefinition)

# Wraps these functions: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/reproject/TileRDDReproject.scala
# All options as above but we also can also accept  `bufferSize` and `matchLayerExtent`
```

The first reproject would be used when working with tiles of desperate incompatible projections, before the tiling step and the second can be used after the tiling step to do a correct reprojection of a tile layer. An example of this logic is captured in our ETL module here: https://github.com/locationtech/geotrellis/blob/master/spark-etl/src/main/scala/geotrellis/spark/etl/Etl.scala#L150-L180

PR #85, opened 2017-03-30
Enable All Tests On Travis

jamesmcclain:


geopyspark/tests/base_test_class.py
```diff
@@ -4,4 +4,4 @@
 
 
 class BaseTestClass(unittest.TestCase):
-    geopysc = GeoPyContext(master="local[*]", appName="test")
+    geopysc = GeoPyContext(master="local[1]", appName="test")
```
jamesmcclain: This may be a point of discussion.  The tests appeared to be failing on Travis due to resource constraints.  I did some tests earlier, and running this with `local[32]` would spawn as many as 18 unique (gateway) JVMs.

jbouffard: Is this behavior only seen on Travis? If so, then we could just have a condition where it creates a `GeoPyContext` based its environment.

jamesmcclain: Right, I thought about that, but it also slightly speeds up the tests locally.

Also, the tests are probably going to be run on Travis more consistently than locally, so it is arguable whether different behavior on Travis versus local are useful (that is one questions that I want to open up for discussion).

jbouffard: Does it? Do you think that trend will continue as we add more tests?

Personally, I run a lot of tests locally, but getting them to run on Travis well is also important. I think it might be best to keep it at certain level for both local and Travis for now, but then check on the performance as the tests increase/become more complicated.

What is your opinion @echeipesh?

jamesmcclain:
I changed this so that it one executor is used on travis, and the previous behavior is maintained everywhere else.

PR #86, opened 2017-03-30
Metadata Refactor

jbouffard:
This Pr is based another, open Pr #84 . This Pr seeks to add a new way of collecting metadata through using `FloatingLayoutScheme`. In addition to this new method, the metadata collecting code has been cleaned up and a few function names have been changes.

PR #87, opened 2017-03-30
Geopycontext Refactor

### LINKED ISSUE #83: Refactor GeoPyContext to Hide the Various Property Methods
There currently exists various `property` methods in `GeoPyContext` that act as a bridge between the backend, scala classes/object and python. However, the user has complete access to these methods when they really shouldn't; as there is no reason as to why they'd need to use them. Thus, it'd be best to just hid them from the user.

jbouffard:
This Pr is based on another, open Pr #86 . This Pr seeks to make certain methods of `GeoPyContext` private. The main reason for doing this is these methods act as a bridge between scala classes/objects and python; Thus, the user has no reason to use them and leaving them exposed could cause potential issues.

This Pr fixes #83

### DIFF
--- a/geopyspark/geopycontext.py
+++ b/geopyspark/geopycontext.py
@@ -22,51 +22,51 @@ class GeoPyContext(object):
         self.avroregistry = AvroRegistry()
 
     @property
-    def schema_producer(self):
+    def _schema_producer(self):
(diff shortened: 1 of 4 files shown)


jbouffard:

PR #89, opened 2017-04-04
Add Support For Cost-Distance and Stitch

jamesmcclain:
Contains Cost-Distance and Stitch operations.

### DIFF
--- a/geopyspark/geopyspark_utils.py
+++ b/geopyspark/geopyspark_utils.py
@@ -47,6 +47,7 @@ def setup_environment():
     os.environ["PYSPARK_DRIVER_PYTHON"] = "python3"
     os.environ["PYSPARK_SUBMIT_ARGS"] = "--jars {} \
             --conf spark.ui.enabled=false \
+            --conf spark.serializer=org.apache.spark.serializer.KryoSerializer \
             --driver-memory 8G \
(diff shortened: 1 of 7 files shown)


geopyspark/geopyspark_utils.py
```diff
@@ -47,6 +47,7 @@ def setup_environment():
     os.environ["PYSPARK_DRIVER_PYTHON"] = "python3"
     os.environ["PYSPARK_SUBMIT_ARGS"] = "--jars {} \
             --conf spark.ui.enabled=false \
+            --conf spark.serializer=org.apache.spark.serializer.KryoSerializer \
```
jamesmcclain: Possibly of interest.  This is necessary because Python `lists` become `java.util.ArrayList`s which are evidently not serializable by the default mechanism.

echeipesh: Good to know, thanks for calling that out

jbouffard:
@jamesmcclain Is this branch rebased on top of the current master? Because your PR doesn't have the `focal` operation. Which is causing the merge conflicts.

jamesmcclain:
No, both this and focal were based on top of what was (at the time of creation) current master.  I just rebased, will push changes when tests finish.

PR #92, opened 2017-04-05
Add Support For Mask Operation

jamesmcclain:
Support has been added for masking operations using both the GeoTrellis machinery as well as standard Python/rasterio machinery.


jbouffard:
:+1:

jamesmcclain:
Thanks

PR #94, opened 2017-04-06
TileRasterRDD Refactor

jbouffard:
This Pr seeks to add `zoomLevel: Option[Int]` as one of the parameters for the `TileRasterRDD[_]` classes. By doing this, it will be easier to manage each `tileLayer` both in python and in scala.

### DIFF
--- a/geopyspark/rdd.py
+++ b/geopyspark/rdd.py
@@ -57,9 +57,19 @@ class TiledRasterRDD(object):
         self.rdd_type = rdd_type
         self.srdd = srdd
 
+    @property
     def layer_metadata(self):
         """Layer metadata associated with this layer"""
         return json.loads(self.srdd.layerMetadata())
+
+    @property
+    def zoom_level(self):
+        zoom = self.srdd.getZoom()
+
+        if zoom >= 0:
+            return zoom
+        else:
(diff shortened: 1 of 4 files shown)


geopyspark/rdd.py
```diff
@@ -57,10 +57,20 @@ def __init__(self, geopysc, rdd_type, srdd):
...
+    @property
     def layer_metadata(self):
         """Layer metadata associated with this layer"""
         return json.loads(self.srdd.layerMetadata())
 
+    @property
+    def zoom_level(self):
+        zoom = self.srdd.getZoom()
+
+        if zoom >= 0:
```
jbouffard: @echeipesh What do you think of sending a value back to python even when a scala method failed so an `Exception` can be raised on the python side?

jbouffard:

geopyspark/rdd.py
echeipesh: I think its good idea to raise errors on python side as much as possible but in this case we can just return something that gets marshaled to python `None` with less surprise

```scala
  def getZoom: Integer = zoomLevel match {
    case Some(z) => new Integer(z)
    case None => null
  }
```

echeipesh:
Superseded by https://github.com/locationtech-labs/geopyspark/pull/113

PR #95, opened 2017-04-06
Constants Refactor

jbouffard:
This Pr is based on another, open Pr #94 . This Pr simply moves `geopyspark.geotrellis.constants` to `geopyspark.constants`. The reasoning behind this is because `rdd.py` uses constants, and before the move had to access a submodule to retrieve them. Thus, to keep things more in-line with the library's structure, `constants.py` was moved up a directory.

### DIFF
--- a/geopyspark/rdd.py
+++ b/geopyspark/rdd.py
@@ -57,9 +57,19 @@ class TiledRasterRDD(object):
         self.rdd_type = rdd_type
         self.srdd = srdd
 
+    @property
     def layer_metadata(self):
         """Layer metadata associated with this layer"""
         return json.loads(self.srdd.layerMetadata())
+
+    @property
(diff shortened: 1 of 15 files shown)


geopyspark/rdd.py
```diff
@@ -57,10 +59,20 @@ def __init__(self, geopysc, rdd_type, srdd):
         self.rdd_type = rdd_type
         self.srdd = srdd
 
+    @property
```
echeipesh: What is the purpose of the `@property` here? I'm not sure if there is extra magic that is happening?

echeipesh: Ah, it allows to call the method without the parens.

echeipesh:
Superseded by https://github.com/locationtech-labs/geopyspark/pull/113

PR #117, opened 2017-04-13
Documentation

jbouffard:
This Pr is based on another, open Pr #116 . This PR updates and expands all python docstrings in addition to creating various written guides and tutorials on how to use GeoPySpark.

This Pr resolves #73

PR #119, opened 2017-04-19
Reclassify

jbouffard:
This Pr is based on another, open #117 . This Pr adds the `reclassify` method to the `RasterRDD` and the `TiledRasterRDD` classes.


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
```diff
@@ -56,6 +57,32 @@ abstract class TileRDD[K: ClassTag] {
...
+
+  def reclassify(
+    intMap: java.util.Map[Int, Int]
+  ): TileRDD[_] = {
+    val scalaMap = intMap.asScala.toMap
+
+    reclassify(scalaMap.keySet, BreakMap(scalaMap))
+  }
+
+  def reclassifyDouble(
```
echeipesh: This doesn't really work for doubles since you can't enumerate a floating point range to perform the substitution here. You really need a concept of a range with a starting and ending point.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
```diff
@@ -56,6 +57,32 @@ abstract class TileRDD[K: ClassTag] {
...
+    doubleMap: java.util.Map[Double, Double]
+  ): TileRDD[_] = {
```
echeipesh: This doesn't really work for doubles since you can't enumerate a floating point range to perform the substitution here. You really need a concept of a range with a starting and ending point.

rossbernet: I think most of this got settled through in person discussion.

But to respond, and make sure we are on the same page:

1. doesn't seem necessary. And it seems like `BreakMap` makes more sense, and doens't allow 'sparce' reclass anyway.

2. Seems to me that following convention makes sense, but is it possible to allow the ability to override it with increased specificity? (perhaps this would look like default inclusive on the left, and exclusive on the right, but using a bracket instead of a paren overrides it? Maybe thats confusing.) Looks like Jake updated this already. Is it possible to make overriding the default convention an optional arg?)


(0:5 -> 1)
(5:10 ->2)

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
```diff
@@ -129,8 +167,48 @@ class ProjectedRasterRDD(val rdd: RDD[(ProjectedExtent, MultibandTile)]) extends
...
+        for (x <- 0 until count) {
+          val band = m.band(x)
+          tiles(x) = band.map(i => breakMap.apply(i))
+        }
+
+        (k, MultibandTile(tiles))
+      }
+    }
+
+    ProjectedRasterRDD(reclassifiedRDD)
```
echeipesh: Can this be implemented in the abstract class? We don't have to do anything to the key, so we should be able to just pass it through.

jbouffard: Yeah, it can. I'll make the needed changes soon.

jbouffard: Just updated this Pr with that refactor.

PR #121, opened 2017-04-22
Local Operations

jbouffard:
This Pr adds local operations for `TiledRasterRDD`s. The operations added are: `+`, `-`, `*`, and `/`. They can be performed with `int`s, `float`s, and other `TiledRasterRDD`s.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
@@ -135,8 +135,14 @@ abstract class TiledRasterRDD[K: SpatialComponent: AvroRecordCodec: JsonFormat: 
   def localSubtract(i: Int): TiledRasterRDD[_] =
     localSubtract(rdd.map { x => (x._1, MultibandTile(x._2.bands.map { y => y - i })) })
 
+  def reverseLocalSubtract(i: Int): TiledRasterRDD[_] =
+    localSubtract(rdd.map { x => (x._1, MultibandTile(x._2.bands.map { y => y.-:(i) })) })
+
   def localSubtract(d: Double): TiledRasterRDD[_] =
     localSubtract(rdd.map { x => (x._1, MultibandTile(x._2.bands.map { y => y - d })) })
+
+  def reverseLocalSubtract(d: Double): TiledRasterRDD[_] =
+    localSubtract(rdd.map { x => (x._1, MultibandTile(x._2.bands.map { y => y.-:(d) })) })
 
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
```diff
@@ -241,6 +318,18 @@ class SpatialTiledRasterRDD(
...
+    SpatialTiledRasterRDD(zoomLevel, MultibandTileLayerRDD(result, rdd.metadata))
+
+  def localSubtract(result: RDD[(SpatialKey, MultibandTile)]): TiledRasterRDD[SpatialKey] =
+    SpatialTiledRasterRDD(zoomLevel, MultibandTileLayerRDD(result, rdd.metadata))
+
+  def localMultiply(result: RDD[(SpatialKey, MultibandTile)]): TiledRasterRDD[SpatialKey] =
+    SpatialTiledRasterRDD(zoomLevel, MultibandTileLayerRDD(result, rdd.metadata))
+
+  def localDivide(result: RDD[(SpatialKey, MultibandTile)]): TiledRasterRDD[SpatialKey] =
+    SpatialTiledRasterRDD(zoomLevel, MultibandTileLayerRDD(result, rdd.metadata))
```
echeipesh: All of these have the same signature and are just constructor wrappers, maybe `def withRDD(rdd: ..)`

jbouffard: Just resolved this.

geopyspark/geotrellis/rdd.py
```diff
@@ -478,3 +478,39 @@ def reclassify(self, value_map, data_type, boundary_strategy=LESSTHANOREQUALTO):
...
         srdd = _reclassify(self.srdd, value_map, data_type, boundary_strategy)
         return TiledRasterRDD(self.geopysc, self.rdd_type, srdd)
+
+    def _process_operation(self, value, operation):
+        if isinstance(value, int) or isinstance(value, float):
+            srdd = operation(value)
+        elif isinstance(value, TiledRasterRDD):
+            if self.rdd_type != value.rdd_type:
+                raise ValueError("Both TiledRasterRDDs need to have the same rdd_type")
+            srdd = operation(value.srdd)
```
echeipesh: Should also check if the layout if for tiles being the same size, if they're not the local operations will throw during job execution anyway.

echeipesh: Actually since we have the metadata we should really check that the CRS and layouts themselves are the same. The only tricky part is that CRS checking needs to be clever since they're are multiple ways to specify equivalent CRS.

PR #123, opened 2017-04-24
Pre Publish 0.1

### LINKED ISSUE #91: Better Error Handling
When something fails in GeoPySpark, the resulting stack trace is almost always very long and contains information that is not useful for debugging. Better error handling should be introduced so that the resulting stack trace will contain more helpful and terse information. Making it easy to diagnose an issue.

jbouffard:
This PR seeks to tie up any lose ends before we publish to 0.1. This should be the last PR to merged before the release.

This Pr resolves #91


echeipesh:
Aside from changes to docs this seems mostly style and updating remaining tests, so it looks like its good to merge whenever the tests are updated.

PR #131, opened 2017-04-26
[WIP] Assembly From GitHub

jbouffard:
This PR makes it so that when GeoPySpark is installed the jar is taken from GitHub instead of being assembled locally.


jbouffard:
Superseded by #145

PR #140, opened 2017-04-27
Added flag to reclassify to allow NODATA to be remapped

### LINKED ISSUE #136: Reclassify Method Does Not Reclassify NoData Correctly
The `reclassify` method does not reclassify `nan` or `NODATAINT` values and instead just leaves them how they were. This should be resolved before v0.1.0 is released. In addition to correcting this issue, a unit test should be added that covers this instance for both `int`s and `float`s.

jpolchlo:
There was no good way to replace nodata values in a raster.  This PR adds the `replace_nodata_with` flag to allow nodata to be dealt with by `reclassify`.

Fixes #136 

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
@@ -64,13 +64,18 @@ abstract class TileRDD[K: ClassTag] {
 
   def reclassify(
     intMap: java.util.Map[Int, Int],
-    boundaryType: String
+    boundaryType: String,
+    ignoreNoData: Boolean = false
   ): TileRDD[_] = {
     val scalaMap = intMap.asScala.toMap
 
     val boundary = getBoundary(boundaryType)
     val mapStrategy = new MapStrategy(boundary, NODATA, NODATA, false)
-    val breakMap = new BreakMap(scalaMap, mapStrategy, { i: Int => isNoData(i) })
+    val breakMap = 
+      if (ignoreNoData)
+        new BreakMap(scalaMap, mapStrategy, { i: Int => false })
+      else
+        new BreakMap(scalaMap, mapStrategy, { i: Int => isNoData(i) })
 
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
```diff
@@ -89,13 +94,18 @@ abstract class TileRDD[K: ClassTag] {
 
   def reclassifyDouble(
     doubleMap: java.util.Map[Double, Double],
-    boundaryType: String
+    boundaryType: String,
+    ignoreNoData: Boolean = false
```
jbouffard: Since the default value is already set for `ignoreNoData` in python, is there a point to having a default here?

jpolchlo:
This whole PR may be moot.  BreakMap won't do the right thing for floats; it can't search for NaN in a meaningful way, since all tests against NaN return `false`.  It may be required to include an operation for "unmasking" a raster—that is, removing nodata values as a separate step.

jbouffard:
Could you rebase on top of master and see if maybe that helps?

geopyspark/geotrellis/rdd.py
```diff
@@ -527,19 +539,24 @@ def reclassify(self, value_map, data_type, boundary_strategy=LESSTHANOREQUALTO):
...
         Returns:
-            :class:`~geopyspark.geotrellis.rdd.TiledRasterRDD`
+            :class:`~geopyspark.geotrellis.rdd.RasterRDD`
         """
 
-        srdd = _reclassify(self.srdd, value_map, data_type, boundary_strategy)
-        return TiledRasterRDD(self.geopysc, self.rdd_type, srdd)
+        srdd = _reclassify(self.srdd, value_map, data_type, boundary_strategy, replace_nodata_with)
+
+        return RasterRDD(self.geopysc, self.rdd_type, srdd)
```
jbouffard: Why does this return a `RasterRDD`?

jbouffard:
:+1:

PR #143, opened 2017-04-28
Fix Makefile

jamesmcclain:
- The wheel should depend on the jar
   - The jar should depend on the *.scala, not *.jar

The second of those was most-likely the source of the circular dependency warning that would occasionally pop up.

Signed-off-by: James McClain <jmcclain@azavea.com>


jbouffard:
:+1:

PR #144, opened 2017-04-28
Tile Lookup

jamesmcclain:
Given a column and row, gives the ability to retrieve a spatial tile from an RDD.  Also contains change that seems to help tests on Travis.

### DIFF
--- a/geopyspark/geotrellis/rdd.py
+++ b/geopyspark/geotrellis/rdd.py
@@ -401,6 +401,25 @@ class TiledRasterRDD(object):
 
         return TiledRasterRDD(self.geopysc, self.rdd_type, srdd)
 
+    def lookup(self, col, row):
+        """Return the value(s) in the image of a particular SpatialKey (given by col and row)
+
+        Args:
+            col: The SpatialKey column
+            row: The SpatialKey row
+
+        Returns: An array of numpy arrays (the tiles)
+        """
+        if self.rdd_type != SPATIAL:
+            raise ValueError("Only TiledRasterRDDs with a rdd_type of Spatial can use lookup()")
+
+        tup = self.srdd.lookup(col, row)
+        array_of_tiles = tup._1()
+        schema = tup._2()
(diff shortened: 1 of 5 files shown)


geopyspark/geotrellis/rdd.py
```diff
@@ -401,6 +401,25 @@ def reproject(self, target_crs, extent=None, layout=None, scheme=FLOAT, tile_siz
 
         return TiledRasterRDD(self.geopysc, self.rdd_type, srdd)
 
+    def lookup(self, col, row):
+        """Return the value(s) in the image of a particular SpatialKey (given by col and row)
+
+        Args:
+            col: The SpatialKey column
+            row: The SpatialKey row
```
jbouffard: It'd be nice if the docstrings were formatted similarly to the other docstrings in the file.

For instance, the `Args` would be formatted like:
```python
   col (int): The SpatialKey column
```

geopyspark/geotrellis/rdd.py
```diff
@@ -401,6 +401,25 @@ def reproject(self, target_crs, extent=None, layout=None, scheme=FLOAT, tile_siz
...
+
+        Returns: An array of numpy arrays (the tiles)
+        """
+        if self.rdd_type != SPATIAL:
+            raise ValueError("Only TiledRasterRDDs with a rdd_type of Spatial can use lookup()")
+
```
jbouffard: We can get the `Bounds` from the `layer_metadata`, so it might be good to check to see if the `col` and `row` given fall within the `Bounds`. This would return a more readable and understandable stack trace if an Exception is raised.

geopyspark/geotrellis/rdd.py
jamesmcclain: Okay, updated.

geopyspark/geotrellis/rdd.py
jamesmcclain: Okay, updated.

PR #145, opened 2017-04-28
Setup Script for GeoPySpark

jbouffard:
This PR adds a setup script for GeoPySpark that lets the user set a location for the backend jar to be downloaded (if they wish to set a different location).

This PR replaces #131


jbouffard:
@echeipesh @jpolchlo @jamesmcclain Could one of you guys take a look at this PR and try it out if you get the chance? There's instructions on how to install the script in the `REAME`.

jbouffard:
@echeipesh Could you review this when you get the chance?

To get GeoPySpark from PyPiTest, run this:
`pip install --user -i https://testpypi.python.org/pypi geopyspark`

echeipesh:
I was able to get it installed. Although for me it installed `geopyspark` into `/Users/eugene/Library/Python/3.6/bin` which was not on my `PATH` and I had to go looking for it. Is this a misconfiguration on my system and something that can expect to be true otherwise?

But once found it downloaded the jar into `jars` correctly.

jbouffard:
Could you show me your `PATH`? It could have possibly been hidden in there. Where do they usually install to?  It should install to whatever place is set on your computer.

echeipesh:
My path is pretty naked: `/usr/local/bin /usr/bin /bin /usr/sbin /sbin /opt/X11/bin /Applications/Wireshark.app/Contents/MacOS`

I'm not sure I've seen python installing executable programs before, but I guess this is how they're supposed to do it. Going to chuck this to misconfiguration on my part. Since I don't have a suggestion on how better to handle this bootstrap I think this PR is 👍  from me.

PR #146, opened 2017-04-28
Allow loading multiple GeoTIFFs via `get`

jpolchlo:
This PR introduces the ability to supply a list of URIs to `geopyspark.geotrellis.geotiff_rdd.get`.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/geotiff/GeoTiffRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/geotiff/GeoTiffRDD.scala
@@ -81,37 +81,6 @@ object GeoTiffRDD {
     }
   }
 
-  def get(
-    sc: SparkContext,
-    keyType: String,
-    path: String
-  ): RasterRDD[_] = {
-    val uri = new URI(path)
-
-    uri.getScheme match {
-      case S3 =>
-        getS3GeoTiffRDD(sc, keyType, uri, S3GeoTiffRDDOptions.default)
-      case _ =>
-        getHadoopGeoTiffRDD(sc, keyType, new Path(path), HadoopGeoTiffRDDOptions.default)
-    }
-  }
-
-  def get(
-    sc: SparkContext,
-    keyType: String,
-    path: String,
-    options: java.util.Map[String, Any]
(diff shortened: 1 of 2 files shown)


jpolchlo:
No idea how to write a test for this.  I'll take suggestions, but for now removing the WIP tag.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/geotiff/GeoTiffRDD.scala
```diff
@@ -106,10 +108,63 @@ object GeoTiffRDD {
...
+        }
+      }
+  }
+
+  def get_many(
+    sc: SparkContext,
+    keyType: String,
+    paths: java.util.List[String],
+    options: java.util.Map[String, Any]
+  ): RasterRDD[_] = {
```
echeipesh: This should be the only function we expose with python wrapper putting the `uri` in a list, if its not and providing empty param list for options. Otherwise this produces too much code to herd that basically does the same thing.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/geotiff/GeoTiffRDD.scala
```diff
@@ -79,35 +81,57 @@ object GeoTiffRDD {
     }
   }
 
-  def get(
+  def get_many(
```
echeipesh: To nit-pick the name this should still be a `get`

echeipesh: Also I think its safe only to have the version that requires options, which could be empty. Am I missing anything ?

geopyspark/geotrellis/geotiff_rdd.py
```diff
@@ -72,13 +73,24 @@ def get(geopysc,
         options = kwargs
 
     if options:
-        srdd = geotiff_rdd.get(geopysc.sc,
-                               key,
-                               uri,
-                               options)
+        if type(uri) == list:
```
jbouffard: It might be better to use `isinstance(uri, list)` since that would return `True` for `list`s and any of its subclasses.

PR #163, opened 2017-05-03
to_tiled_rdd Method

### LINKED ISSUE #151: Add RasterRDD.to_tiled_layer
The standard workflow is to collect the metadata from an instance of `RasterRDD` and then tile it to to that layout. We made the API design choice is that we always collect the "floating" layout, that matches the input raster resolution and just cuts it to some tile grid of desired tile size (default: 256x256).

In practice the default flow is:
```
metadata = raster_rrd.collect_metadata()
tiled_raster_rdd = rdd.tile_to_layout(metadata)
```

This pattern carries over from GeoTrellis core and the intent is to be explicit when you perform a spark action (`collect_metadata`) and to provide opportunity to override the parameters in `tile_to_layout` step.

However in python API this seems overly verbose and there begs to be a function that combines the two steps into one function that collects the metadata and uses it to generate an instance of `TiledRasterRDD`. This function would have all the parameters of the `collect_metadata` but instead of returning the metadata it can generate the actual "layer" object to be worked with.

jbouffard:
This PR adds the method, `to_tiled_rdd` that combined `collect_metadata` and `tile_to_layout` into one step.

This PR resolves #151

### DIFF
--- a/geopyspark/geotrellis/rdd.py
+++ b/geopyspark/geotrellis/rdd.py
@@ -120,8 +120,8 @@ class RasterRDD(object):
         ser = self.geopysc.create_tuple_serializer(result._2(), value_type=TILE)
         return self.geopysc.create_python_rdd(result._1(), ser)
 
-    def to_tiled_raster_rdd(self, extent=None, layout=None, crs=None, tile_size=256,
-                            resample_method=NEARESTNEIGHBOR):
+    def to_tiled_rdd(self, extent=None, layout=None, crs=None, tile_size=256,
+                     resample_method=NEARESTNEIGHBOR):
         """Converts this ``RasterRDD`` to a ``TiledRasterRDD``.
 
         This method combines :meth:`~geopyspark.geotrellis.rdd.RasterRDD.collect_metadata` and
--- a/geopyspark/geotrellis/rdd.py
+++ b/geopyspark/geotrellis/rdd.py
@@ -120,8 +120,8 @@ class RasterRDD(object):
         ser = self.geopysc.create_tuple_serializer(result._2(), value_type=TILE)
         return self.geopysc.create_python_rdd(result._1(), ser)
 
-    def to_tiled_rdd(self, extent=None, layout=None, crs=None, tile_size=256,
-                     resample_method=NEARESTNEIGHBOR):
+    def to_tiled_layer(self, extent=None, layout=None, crs=None, tile_size=256,
+                       resample_method=NEARESTNEIGHBOR):
         """Converts this ``RasterRDD`` to a ``TiledRasterRDD``.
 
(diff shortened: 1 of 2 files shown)


geopyspark/geotrellis/rdd.py
```diff
@@ -120,6 +120,32 @@ def to_numpy_rdd(self):
         ser = self.geopysc.create_tuple_serializer(result._2(), value_type=TILE)
         return self.geopysc.create_python_rdd(result._1(), ser)
 
+    def to_tiled_rdd(self, extent=None, layout=None, crs=None, tile_size=256,
```
echeipesh: do you think `to_tiled_rdd` is more consistent or clearer than `to_tiled_layer` ?

echeipesh: @rossbernet @jpolchlo tag team to continue naming discussion.

jbouffard: I think so. The reason being is that I try not to refer `TiledRasterRDD`s as `layer`s for the sake of simplicity (though there are a few inconsistencies in the docs). Plus, it matches the return type more closely in terms of name.

rossbernet: @jpolchlo, @echeipesh and I talked about what information is worth trying to hide from the user - or another way, what information and to what level of detail does the user need to understand underlying mechanics of GPS to be able to use it. 

I think it might be worth thinking about 3 hypothetical users: 

Tier 1 - Someone who has very little programming experience. They are interested in learning more. They have a decent GIS background, and they are trying to take it to the next level because they want to process big raster data sets on the web. 

tier 2 - Okay programmer. Okay GIS background. 

tier 3 - Excellent programmer - understand big data. New to GIS.


Tier 1 is the one I think we need to be particularly careful of not intimidating. Tiers 1 and 2 will be able hack at stuff so long as the documention is there. I hope the same is true for Tier 1 so long as the documentation is good enough.

But. Question for Tier 1 in particular:

1. Is it possible to hide or minimize the fact that they are using Spark? (Is that even a worthwhile goal?) How much do they need to understand what an RDD is? 

Basically - minimizing new concecpts / acronyms is a worthy goal if they don't need to understand them to gain basic benefits from GPS.

On that note - `to_tiled_layer` might make more sense if they dont really need to understand what an RDD is, because a layer (as defined by a set of rasters) is more intuitive.

jbouffard: Okay, I'll change it then.

Though, because we reference both Spark and RDDs throughout the library I don't think it's really possible to hide those terms from the user. In addition, having knowledge of Spark and its associated terms would allow for the user to better understand the advantages/disadvantages of using GeoPySpark vs another geospatial library.

jbouffard:
@echeipesh This has been updated and is ready for review.

PR #166, opened 2017-05-04
Convert Raster Type

### LINKED ISSUE #152: No clear way to convert the type of a raster
There should be a means to coerce a raster from `int` to `double` and vice-versa.  Currently there is no clear way to accomplish this.

jbouffard:
This PR gives the user the ability to convert the underlying `Raster` data type to either `int` or `float` (`Double` in Scala).

Note: It is not possible to recreate the `to_int` and `to_float` methods via `int(raster_rdd)` and `float(raster_rdd)`. This is because `def __int__(self, x)` and `def __float__(self, x)` expects an `int` and `float` to be returned, respectively.

This PR resolves #152

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
@@ -123,8 +123,16 @@ abstract class TileRDD[K: ClassTag] {
     }
   }
 
+  def toInt: TileRDD[_] =
+    toInt(rdd.map { x => (x._1, x._2.convert(IntCellType)) })
+
+  def toDouble: TileRDD[_] =
+    toDouble(rdd.map { x => (x._1, x._2.convert(DoubleCellType)) })
+
   protected def reclassify(reclassifiedRDD: RDD[(K, MultibandTile)]): TileRDD[_]
   protected def reclassifyDouble(reclassifiedRDD: RDD[(K, MultibandTile)]): TileRDD[_]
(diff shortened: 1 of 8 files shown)


jbouffard:
@echeipesh I'm not sure why Travis was never kicked off for this PR. Would you want me to resubmit it?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
```diff
@@ -123,8 +123,16 @@ abstract class TileRDD[K: ClassTag] {
     }
   }
 
+  def toInt: TileRDD[_] =
+    toInt(rdd.map { x => (x._1, x._2.convert(IntCellType)) })
+
+  def toDouble: TileRDD[_] =
+    toDouble(rdd.map { x => (x._1, x._2.convert(DoubleCellType)) })
```
echeipesh: These aren't enough unfortunately. Aside from converting between floating and integer values you may also need to control the width of your data types in order to save storage/memory.
So we need to mirror this function: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/MultibandTileLayerRDDMethods.scala#L28

Thankfully there is a mapping from String => CellType: https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/CellType.scala#L412

jbouffard:
@echeipesh This has been updated and should be good for another review.

lossyrob:
Merging this, there are open issues to track other changes.

PR #167, opened 2017-05-04
Min Max Method

jbouffard:
This PR adds the `get_min_max()` method to `RasterRDD` and `TiledRasterRDD` which returns the min and max value of all of the rasters within the `RDD` as `(float, float)`.


jbouffard:
@jpolchlo Could you please review this?

PR #174, opened 2017-05-04
FastAvro

jbouffard:
This PR uses fastavro to do the serialization/deserialization of data.

PR #176, opened 2017-05-05
Update GeoNotebook

jamesmcclain:
It is no longer necessary to map port 8033 out of the container, all communication is done on port 8000.  Handling of adding layers has been changed and support for removing layers has been added.

This should go in after locationtech-labs/geopyspark#174


jbouffard:
:+1:

PR #181, opened 2017-05-08
Add PNG rendering functionality

jpolchlo:
This PR enables the rendering of tiled rasters to a "tiled png" format.  This format will be required to allow users to control the conversion of raster data to image data for display in a geonotebook.  This will also allow for more straightforward integration with geonotebook (i.e., geonotebook may not need to depend on geopyspark).


jpolchlo:
Initial display is now working using the PngRDD method.  The following code (requires some omitted boilerplate code) will display the TIFF data on the map.

```python
from geopyspark.geotrellis.geotiff_rdd import get
from geopyspark.geotrellis.render import PngRDD

rdd = get(geopysc, SPATIAL, 's3://elevation-tiles-prod/geotiff/7/67/51.tif')
png = PngRDD.makePyramid(rdd.to_tiled_layer(), 'viridis')

from geonotebook.wrappers import RddRasterData
M.add_layer(RddRasterData(png, 'DEM'))```

Errors will be seen at higher zoom levels due to the lack of a common histogram during PNG conversion.  But this is no longer a problem that strictly requires iteration on GeoNotebook code.  However, https://github.com/jamesmcclain/geonotebook/pull/5 is the relevant resource for the required changes to that codebase.

[Note that the geonotebook code is now carrying a template for how to visually debug problems in the tile server.  Tiles which have no data in the catalog will display a procedurally-generated tile image which can be modified to display whatever information might be useful.  It remains to allow the passing of a debug flag from the client to selectively activate and deactivate this feature.]

**Comment by jamesmcclain** [comment]
Okay, I will take a look.

File: docker/config/patch.diff [hunk 1]
```diff
@@ -169,45 +170,51 @@ index 0000000..2c7fceb
 +
 +    def ingest(self, data, name, **kwargs):
 +        from geopyspark.geotrellis.rdd import RasterRDD, TiledRasterRDD
-+        from geopyspark.geotrellis.constants import ZOOM
++        from geopyspark.geotrellis.render import PngRDD
++        # from geopyspark.geotrellis.constants import ZOOM
```
jbouffard: If the commented out stuff isn't needed, can it be removed?

docker/config/patch.diff
```diff
@@ -311,30 +329,28 @@ index f95e8af..6a0e86b 100644
...
 +class RddRasterData(object):
 +
-+    def __init__(self, rdd, name=None, remapper=None):
++    def __init__(self, rdd, name=None, rampname="Viridis"):
 +        from geopyspark.geotrellis.rdd import RasterRDD, TiledRasterRDD
++        from geopyspark.geotrellis.render import PngRDD
 +
-+        if not (isinstance(rdd, RasterRDD) or isinstance(rdd, TiledRasterRDD)):
++        if not (isinstance(rdd, RasterRDD) or isinstance(rdd, TiledRasterRDD) or isinstance(rdd, PngRDD)):
 +            raise Exception
```
jbouffard: Same as the above comment.

geopyspark/tests/png_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class PngRddTest(BaseTestClass):
+
+    @pytest.fixture(autouse=True)
+    def tearDown(self):
+        yield
+        BaseTestClass.geopysc.pysc._gateway.close()
+
+    def test_if_working(self):
```
jbouffard: Are there any other tests that can be done to ensure that `PngRDD` works?

geopyspark/geotrellis/render.py
```diff
@@ -0,0 +1,56 @@
+from geopyspark.geotrellis.constants import NEARESTNEIGHBOR, ZOOM
+
+class PngRDD(object):
```
jbouffard: Could you please add docstrings to this file? https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/rdd.py Has good examples of the docstring style.

geopyspark/geotrellis/render.py
```diff
@@ -0,0 +1,56 @@
...
+        level0 = pyramid[0]
+        self.geopysc = level0.geopysc
+        self.rdd_type = level0.rdd_type
+        self.layer_metadata = list(map(lambda lev: lev.layer_metadata, pyramid))
+        self.max_zoom = level0.zoom_level
+        self.pngpyramid = list(map(lambda layer: self.geopysc._jvm.geopyspark.geotrellis.PngRDD.asSingleband(layer.srdd, rampname), pyramid))
+
+    @classmethod
+    def makePyramid(cls, tiledrdd, rampname, start_zoom=None, end_zoom=0, resamplemethod=NEARESTNEIGHBOR):
+        reprojected = tiledrdd.reproject("EPSG:3857", scheme=ZOOM)
```
jbouffard: 1. `resamplemethod` should be changed to `resample_method` so that it's consistent with the rest of library, and so that it's consistent with Python's naming conventions.

2. There should be a check to see if `resample_method` is a known resampling method. Like https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/rdd.py#L202

geopyspark/geotrellis/render.py
```diff
@@ -0,0 +1,56 @@
+from geopyspark.geotrellis.constants import NEARESTNEIGHBOR, ZOOM
+
+class PngRDD(object):
+    def __init__(self, pyramid, rampname):
```
jbouffard: `rampname` should be changed to `ramp_name` to fit Python's naming conventions.

docker/config/patch.diff
```diff
@@ -169,45 +170,51 @@ index 0000000..2c7fceb
...
++            png = PngRDD.makePyramid(laid_out, data.rampname)
 +        elif isinstance(rdd, TiledRasterRDD):
 +            laid_out = rdd
-+            reprojected = laid_out.reproject("EPSG:3857", scheme=ZOOM)
++            # reprojected = laid_out.reproject("EPSG:3857", scheme=ZOOM)
++            png = PngRDD.makePyramid(laid_out, data.rampname)
++        elif isinstance(rdd, PngRDD):
++            png = rdd
 +        else:
 +            raise Exception
```
jbouffard: This should raise a `TypeError` with a description of the error, not just a plain `Exception`.

geopyspark/tests/png_test.py
jpolchlo: If there were a more robust system for color mapping, then yes.  We could pull a tile from a tiled rdd, color map it, pack it into an image, and compare it with the result of the PngRDD lookup.  But we don't currently have the facilities for color mapping that would allow this.  I can create a TODO comment in this file to remind us to add a test when we have the ability to do so?

jbouffard: Yeah, I think that's fine.

PR #197, opened 2017-05-15
[WIP] GeoNotebook demo changes

lossyrob:
There's some core changes that are a little half baked in here:
- Getting hex values as strings from color ramps
- Getting quantile breaks
- repartitioning
- adding sequences of rdds (I only added localAdd)


docker/config/patch.diff
```diff
@@ -155,53 +312,256 @@ index 0000000..029493e
...
++        inproc_server_states = kwargs.pop('inproc_server_states', None)
++        if inproc_server_states is None:
++            raise Exception(
++                "GeoTrellis vis server requires kernel_id as kwarg to ingest!")
++
++        if not "geotrellis" in inproc_server_states:
++            inproc_server_states["geotrellis"] = { "ports" : {} }
++
++        server_port = randint(49152, 65535)
++        while server_port in inproc_server_states['geotrellis']['ports'].values():
```
jamesmcclain: Does this work when there is more than one document open (and/or more than one user)?

echeipesh: I have used this with multiple kernels running at the same time. Not sure what more than one user means right now though.

echeipesh:
Superseded by: https://github.com/locationtech-labs/geopyspark/pull/210

docker/config/patch.diff
jamesmcclain: The collision will only happen with low probability regardless, but I was wondering if the list of used ports is shared between notebooks, in particular notebooks being run by different users.  I was hoping that one of these python web servers would support binding a port then reporting which one it had bound to, thereby ensuring that there is no collision.

PR #203, opened 2017-05-15
Add numPartitions argument to query and read

echeipesh:
The default behavior of using `defaultParallism` is often undesirable.

### DIFF
--- a/geopyspark/geotrellis/catalog.py
+++ b/geopyspark/geotrellis/catalog.py
@@ -167,6 +167,7 @@ def read(geopysc,
          layer_name,
          layer_zoom,
          options=None,
+         numPartitions=None,
          **kwargs):
 
     """Reads a single, zoom layer from a GeoTrellis catalog.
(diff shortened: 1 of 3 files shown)


geopyspark/geotrellis/catalog.py
```diff
@@ -167,6 +167,7 @@ def read(geopysc,
          layer_name,
          layer_zoom,
          options=None,
+         numPartitions=None,
```
jbouffard: Could you update the docstring so that `numPartitions` is included?

geopyspark/geotrellis/catalog.py
```diff
@@ -289,6 +294,7 @@ def query(geopysc,
           time_intervals=None,
           proj_query=None,
           options=None,
+          numPartitions=None,
```
jbouffard: Same as the above comment.

jbouffard:
:+1:

Issue #139, opened 2017-04-27
Expose persist and unpersist methods on RDD wrappers


echeipesh:
When RDD wrapped by TiledRasterRDD is going to be accessed multiple times it makes sense to cache it in the workbook. To that end we should expose the persist/cache methods and make sure to follow the method signatures from python `RDD` class.

PR #204, opened 2017-05-16
Add persist, unpersist, and cache methods to RasterRDD wrappers

echeipesh:
Resolves: https://github.com/locationtech-labs/geopyspark/issues/139

### DIFF
--- a/geopyspark/geotrellis/rdd.py
+++ b/geopyspark/geotrellis/rdd.py
@@ -50,6 +50,14 @@ def _reclassify(srdd, value_map, data_type, boundary_strategy, replace_nodata_wi
 
 
 class RDDWrapper(object):
+    """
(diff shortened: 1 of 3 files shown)


geopyspark/geotrellis/rdd.py
```diff
@@ -48,7 +49,41 @@ def _reclassify(srdd, value_map, data_type, boundary_strategy, replace_nodata_wi
             return srdd.reclassifyDouble(new_dict, boundary_strategy, replace_nodata_with)
 
 
-class RasterRDD(object):
+class RDDWrapper(object):
```
jbouffard: Could you also add a docstring for this class? Something like this https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/rdd.py#L52

jbouffard:
:+1:

PR #205, opened 2017-05-16
Polygonal Summary Methods

jbouffard:
This PR adds the basic `PolygonalSummary` methods to the `TiledRasterRDD`. Some of the methods added are `polygonalSum`, `polygonalMean`, and `polygonalMax` among others.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
@@ -211,6 +211,53 @@ abstract class TiledRasterRDD[K: SpatialComponent: AvroRecordCodec: JsonFormat: 
     withRDD(rdd.convert(CellType.fromName(newType)))
 
   protected def withRDD(result: RDD[(K, MultibandTile)]): TiledRasterRDD[_]
+
+  def singleTileLayerRDD: TileLayerRDD[K] = TileLayerRDD(
+    rdd.map({ case (k, v) => (k, v.band(0)) }),
+    rdd.metadata
+  )
+
+  def polygonalMin(geom: String): Int =
+    WKT.read(geom) match {
+      case poly: Polygon => singleTileLayerRDD.polygonalMin(poly)
+      case multi: MultiPolygon => singleTileLayerRDD.polygonalMin(multi)
+    }
+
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
```diff
@@ -211,6 +211,53 @@ abstract class TiledRasterRDD[K: SpatialComponent: AvroRecordCodec: JsonFormat:
     withRDD(rdd.convert(CellType.fromName(newType)))
 
   protected def withRDD(result: RDD[(K, MultibandTile)]): TiledRasterRDD[_]
+
+  def singleTileLayerRDD: TileLayerRDD[K] = TileLayerRDD(
+    rdd.map({ case (k, v) => (k, v.band(0)) }),
+    rdd.metadata
+  )
```
echeipesh: This is rough, I guess we only have polygonal summaries defined on single-band tiles in geotrellis at the moment.

echeipesh: @jbouffard  This is something we'll have to address soon. Do you think it makes sense to return an array of results to future proof this interface? Currently it would have to return an array of a single item.

jbouffard: Yeah, that's the current state of the polygonal summary methods. Should I make it so that the operation is done per band and then do a final summary for all of the resulting values?

Also many other methods that are restricted to `Singleband` use this method in the backend.

jbouffard: Sorry, just saw your other comment. I think that makes sense, though, we might have to implement that in more methods than just the polygonal ones.

echeipesh: Having a think about it, I think this is good. I imagine future function will be overloaded on the type. I expect that in multiband case you would not always apply the same summary for all the bands. ex:
```py
stats = rdd.polygonalSummary(geom, [MAX, MAX, AVERAGE])
# stats is [float, float, float]
```

PR #206, opened 2017-05-16
Read Metadata from GeoTrellis Layer

### LINKED ISSUE #190: Expose metadata fetching in python
I want to get a layer metadata; in order to do that I have to read the RDD and grab the metadata off of it, even though I won't be using the rdd. We should have a way to get the metadata of a layer without it.

jbouffard:
This PR adds the `read_metadata` function to `catalog`. This will allow the user to read `tileLayerMetadata` from a saved, GeoTrellis layer without having to read in the whole layer.

This PR resolves #190

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
@@ -50,12 +50,6 @@ abstract class LayerReaderWrapper {
     projQuery: String,
     numPartitions: Int
   ): TiledRasterRDD[_]
-
-  def readMetadata(
-    keyType: String,
-    layerName: String,
-    zoom: Int
-  ): String
 }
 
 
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
```diff
@@ -50,6 +50,12 @@ abstract class LayerReaderWrapper {
     projQuery: String,
     numPartitions: Int
   ): TiledRasterRDD[_]
+
+  def readMetadata(
+    keyType: String,
+    layerName: String,
+    zoom: Int
+  ): String
```
echeipesh: This seems like it belongs on `AttributeStoreWrapper` unless there is super compelling reason for it to be here?

geopyspark/geotrellis/catalog.py
```diff
@@ -161,6 +162,50 @@ def _construct_catalog(geopysc, new_uri, options):
                                           value_reader=value_reader,
                                           writer=writer)
 
+def read_metadata(geopysc,
```
echeipesh: Similar function on `TiledRasterRDD` is `def layer_metadata(self)` so this should be `def read_layer_metadata` to keep it consistent.

PR #207, opened 2017-05-16
Provide a Euclidean distance operator

jpolchlo:
This PR provides the ability to convert shapely geometries into a tiled Euclidean distance layer, ready to be added to the map.  An example of usage is as follows:

```python
import shapely
line = shapely.geometry.LineString([(1,1),(2,2)])
tiled = TiledRasterRDD.euclidean_distance(geopysc, line, 'EPSG:4326', 7)
```


geopyspark/geotrellis/rdd.py
```diff
@@ -433,6 +433,14 @@ def rasterize(cls, geopysc, rdd_type, geometry, extent, crs, cols, rows,
 
         return cls(geopysc, rdd_type, srdd)
 
+    @classmethod
+    def euclidean_distance(cls, geopysc, geometry, source_crs, zoom):
```
jbouffard: Please add a docstring for this method.

geopyspark/geotrellis/rdd.py
```diff
@@ -433,6 +433,14 @@ def rasterize(cls, geopysc, rdd_type, geometry, extent, crs, cols, rows,
...
+        if isinstance(source_crs, int):
+            source_crs = str(source_crs)
+
+        srdd = geopysc._jvm.geopyspark.geotrellis.SpatialTiledRasterRDD.euclideanDistance(geopysc.sc, dumps(geometry), source_crs, zoom)
```
jbouffard: I know you said that any geometry can be used, but that certain ones would be very slow. Do you think it'd be worth have a warning message come up if the user is trying to use a suboptimal geometry?

geopyspark/tests/euclidean_distance_test.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+
+import pytest
+
+from shapely.geometry import Point, MultiPoint, LineString
+from geopyspark.tests.base_test_class import BaseTestClass
+from geopyspark.geotrellis.rdd import TiledRasterRDD
+from geopyspark.geotrellis.constants import SPATIAL
+
+import pyproj
```
jbouffard: pyproj will have to be added to `requirements.txt` if this is to pass on Travis. Either that, or an extra line will have to be added to the Travis installation script.

PR #208, opened 2017-05-16
Get Layer Ids from Catalog

jbouffard:
This PR allows the user to get a list of all the `LayerId`s of a GeoTrellis catalog as strings.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
@@ -63,6 +63,9 @@ abstract class FilteringLayerReaderWrapper()
   def attributeStore: AttributeStore
   def layerReader: FilteringLayerReader[LayerId]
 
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderFactory.scala
```diff
@@ -63,6 +63,9 @@ abstract class FilteringLayerReaderWrapper()
   def attributeStore: AttributeStore
   def layerReader: FilteringLayerReader[LayerId]
 
+  def layerIds: Array[String] =
+    attributeStore.layerIds.map { x => x.toString }.toArray
```
echeipesh: Is .toString actually useful here? It will return something like `Layer(name = "asdf", zoom = 3)` which some poor soul would have to parse. Probably we can return a list of dictionaries with `name` and `zoom` keys here.

jbouffard:
@echeipesh Updated this PR

echeipesh:
needs to be rebased now

PR #210, opened 2017-05-17
GeoNotebook demo changes and performance

echeipesh:
Supersedes: https://github.com/locationtech-labs/geopyspark/pull/197

Rebases the PR on master and makes following improvements
- Use caching in the Weighted Overlay notebook
- Make python pyramid functions either preserver or impose partitioning
- PngRDD preserves partitioning of source RDD
- prints replaced by logging
- Bump to GeoTrellis RC5

Partitioning changes allow the .lookup to load a single partition. When an RDD does not have a partitioner the fallback action for .lookup to is to iterate over every record. Not great.

Overall the initial load of a cached RDD (for instance Chatta layers at zoom 12) still takes time but after caching the tile server is very responsive.


lossyrob:
@echeipesh this should be bumped to RC6

PR #211, opened 2017-05-17
Value Reader Check

jbouffard:
This PR makes it so that `read_value` will first check to see if the given `col` and `row` are within the bounds of the layer. If they're not, then `None` is returned. Else then the `Tile` is given back.

This PR will be ready to be reviewed and merged once #206 is merged.


jbouffard:

PR #218, opened 2017-05-18
Save stitched geotiff

lossyrob:
This also bumps up the scala version.

This also adds normalization.

@jpolchlo I also see a code change snuck in that takes away the reproject step PngRDD. I was finding that this was unnecessary (seems like a fair requirement that a PngRDD has to be in the correct projection) and was taking time. Thoughts?


jpolchlo:
The reproject step was as much about knowing the zoom level of a given raster layer.  Reprojecting was the quickest route to a guarantee of having that knowledge.  It's not ideal, but losing this step breaks some things that depend on knowing that zoom level.  We ran into some trouble not having this code in place while working on the demo.

lossyrob:
This PR is getting out of hand, I'm going to have to cherry-pick the good parts and make alternate PRs, including

- saved_stitched
- normalize

jbouffard:
Superseded by #265

PR #233, opened 2017-05-24
Expand development docs section on editable installs

echeipesh:
I've been able to verify these steps on OSX and the development process is glorious.

Combined with:

```
%load_ext autoreload
%autoreload 2
```

It is possible to make changes to GeoNotebook `ingest` methods and see effects on subsequent `M.add_layer` calls.

There doesn't seem to be a graceful way to setup all the environment variables since at this point they're highly system dependent. Getting all the requirements installed can also be a challenge.

Also note that it is important that `jupyter notebook` binds to port `8000`, as `jupyterhub` would, instead of default `8888` for the tile service routes to be discovered correctly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -266,6 +266,32 @@ command:
 
 This will open up the jupyter hub and will allow you to work on your notebooks.
 
+It is also possible to develop with both GeoPySpark and GeoNotebook in editable mode.
+To do so you will need to re-install and re-register GeoNotebook with Jupyter.
+
+.. code:: console
+
+   pip uninstall geonotebook
+   git clone --branch feature/geotrellis https://github.com/geotrellis/geonotebook ~/geonotebook
+   pip install -e ~/geonotebook
+   jupyter serverextension enable --py geonotebook
+   jupyter nbextension enable --py geonotebook
+   make notebook
+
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -266,6 +266,32 @@ command:
...
+.. code:: console
+
+   pip uninstall geonotebook
+   git clone --branch feature/geotrellis https://github.com/geotrellis/geonotebook ~/geonotebook
+   pip install -e ~/geonotebook
+   jupyter serverextension enable --py geonotebook
+   jupyter nbextension enable --py geonotebook
+   make notebook
+
+The default `Geonotebook (Python 3)` kernel will require the following environment variables to be defined:
```
jbouffard: Having a single ` in a rst file makes the word/s italicized. If you wanted to do an in line quote, you need to use ``

jbouffard:
When I try to run the code in a notebook that was started by `make notebook`, I get the following error:

```
The kernel has died, and the automatic restart has failed. It is possible the kernel cannot be restarted. If you are
not able to restart the kernel, you will still be able to save the notebook, but running code will no longer work
until the notebook is reopened.
```

PR #236, opened 2017-05-25
User Definited NoData Value CellType Fix

### LINKED ISSUE #232: CellTypes for User defined NoData values.
Current definition is `FLOAT64UD = "float64ud"`, but parsing requires the actual NoData value after 'ud' marker: https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/CellType.scala#L437

It looks like for these cases instead of constants we should provide functions that map a value to correct string.

jbouffard:
This PR adds a function, `create_no_data_constant` that lets users create a user defined NoData `CellType`.

This PR resolves #232

### DIFF
--- a/geopyspark/geotrellis/constants.py
+++ b/geopyspark/geotrellis/constants.py
@@ -233,26 +233,36 @@ FLOAT32 = "float32"
 """Representes Double Cells with constant NoData values."""
 FLOAT64 = "float64"
 
-"""Representes Byte Cells with user defined NoData values."""
-INT8UD = "int8ud"
+def create_no_data_constant(cell_type, no_data_value):
+    """Creates a ``CellType`` that has a user defined NoData value.
 
-"""Representes UByte Cells with user defined NoData values."""
-UINT8UD = "uint8ud"
+    Cannot be used when the ``cell_type`` is either a boolean or contains raw values.
 
-"""Representes Short Cells with user defined NoData values."""
-INT16UD = "int16ud"
+    Args:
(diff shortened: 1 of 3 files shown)


geopyspark/geotrellis/constants.py
```diff
@@ -233,26 +233,36 @@
...
-FLOAT32UD = "float32ud"
+    if 'bool' in cell_type:
+        raise ValueError("Cannot add user defined types to Bool")
+    elif 'raw' in cell_type:
+        raise ValueError("Cannot add user defined types to raw values")
+    elif cell_type not in CELL_TYPES:
+        raise ValueError(cell_type, "Is not a known CellType")
+    else:
+        no_data_constant = cell_type + "ud" + str(no_data_value)
+        CELL_TYPES.append(no_data_constant)
```
jbouffard: @echeipesh What is your opinion on adding the new ud `CellType` to the list of `CellType` constants? I did this since there's a check [here](https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/rdd.py#L234). It feels a little weird adding something to a list that's supposed to be constant, though. Would it be better to have a separate list that contains just user defined types and the method checks that as well?

echeipesh:
I like the notion but it doesn't quite square with `rdd.layer_metadata` which is going to have `int32ud4` for instance https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/__init__.py#L147

PR #240, opened 2017-06-05
Include hillshade operator in TiledRasterRDD

### LINKED ISSUE #57: Hillshade operation through Python
This is a sample use case to test the ease with which it will be to open a jupyter notebook and use python bindings to calculate hillshade.

The output should be a step by step demo that is as easy to follow as possible.

jpolchlo:
Fixes #57.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
@@ -147,6 +147,13 @@ abstract class TiledRasterRDD[K: SpatialComponent: AvroRecordCodec: JsonFormat: 
     maxDistance: Double
   ): TiledRasterRDD[K]
 
+  def hillshade(sc: SparkContext,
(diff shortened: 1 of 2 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
```diff
@@ -429,6 +436,25 @@ class SpatialTiledRasterRDD(
     SpatialTiledRasterRDD(None, multibandRDD)
   }
 
+  def hillshade(
+    sc: SparkContext,
```
echeipesh: Just noticing this now but we don't need `sc` here, we can and should just pull it from `rdd.sparkContext`

jpolchlo: In which case we should pull it out of the costDistance parameter list as well, since that was my template

jbouffard:
This looks good to me. We could remove the `SparkContext` from `costDistance` in another PR.

Issue #192, opened 2017-05-14
FastAvro throws exception when repartitioning a numpy array and then make a TiledRDD


lossyrob:
Have to investigate this further, but I had an instance where:

```python
nprdd = rdd.to_numpy_rdd().repartition(50).cache()
rc = np.vectorize(reclass)
def mapper(tile):
    arr = tile['data'][0]
    tile['data'] = np.array([rc(arr)])
    return tile
reclassed = nprdd.mapValues(mapper)
gtRdd = TiledRasterRDD.from_numpy_rdd(geopysc, SPATIAL, reclassed, rdd.layer_metadata)
```

eventually failed with a fastavro error about list index out of range, where as

```python
nprdd = rdd.to_numpy_rdd().cache()
rc = np.vectorize(reclass)
def mapper(tile):
    arr = tile['data'][0]
    tile['data'] = np.array([rc(arr)])
    return tile
reclassed = nprdd.mapValues(mapper)
gtRdd = TiledRasterRDD.from_numpy_rdd(geopysc, SPATIAL, reclassed, rdd.layer_metadata)
```

succeeded.

jbouffard:
What was the `reclass` method exactly? And what which data set were you using? Because I was able to run through this example with the repartitioned `RDD` and it worked fine. Though, this was also done locally.

Here's my code:

```python

def reclass(arr):
    arr[100 >= arr] = 100
    return arr

nprdd = tiled_rdd.to_numpy_rdd().repartition(50).chache()
rc = np.vectorize(reclass)
def mapper(tile):
    tile = tile['data'][0]
    tile['data'] = np.array([rec(arr)])
    return tile
reclassed = nprdd.mapValues(mapper)
gtRdd = TiledRasterRDD.from_numpy_rdd(geopysc, SPATIAL, reclassed, tiled_rdd.layer_metadta)
```

jbouffard:
I'm going to tentatively say this issue has been resolved now that we're using ProtoBuf for serialization. However, this could be subject to change if the issue is created from a different source.

Issue #191, opened 2017-05-13
Not writing (or potentially reading) bit rasters correctly


lossyrob:
I ingested a bit raster (the Chattanooga data) successfully, but when I tried to read a value out of the layer it gave an error like this:

```
error: cannot reshape array of size 8192 into shape (256,256) - Traceback (most recent call last):
  File "/home/hadoop/.local/lib/python3.4/site-packages/geonotebook/vis/geotrellis/server.py", line 50, in tile
    return fn(z, x, y)\n  File "/home/hadoop/.local/lib/python3.4/site-packages/geonotebook/vis/geotrellis/server.py", line 89, in tile
    arr = ser.loads(tile._1())[0][\'data\']\n  File "/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/avroserializer.py", line 92, in loads
    return [self.decoding_method(schema_dict)]\n  File "/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/avroregistry.py", line 41, in tile_decoder
    arrs = [cls._tile_decoder(band) for band in bands]\n  File "/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/avroregistry.py", line 41, in <listcomp>
    arrs = [cls._tile_decoder(band) for band in bands]\n  File "/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/avroregistry.py", line 20, in _tile_decoder
    arr = np.array(cells).reshape(schema_dict[\'rows\'], schema_dict[\'cols\'])\nValueError: cannot reshape array of size 8192 into shape (256,256)
```

If I do the ingest like this:

```python
def ingest_layer(layer_name, path):
    rdd = get(geopysc, SPATIAL, path, maxTileSize=512, numPartitions=20)
    metadata = rdd.collect_metadata(tile_size=512)
    metadata['cellType'] = "int8"
    laid_out = rdd.tile_to_layout(metadata)
    reprojected = laid_out.reproject("EPSG:3857", scheme=ZOOM)
    for layer_rdd in reprojected.pyramid(reprojected.zoom_level, 0):
        write(catalog_uri, layer_name, layer_rdd)    
```

it works.

jbouffard:
I actually discovered this bug a few days ago, but failed to report it. Sorry about that. PR #166 fixes this.

jbouffard:
[...]

PR #245, opened 2017-06-14
ProtoBuf Serialization

### LINKED ISSUE #199: Avro decodes int8 tiles as uint8
Should check ushort/short as well.

jbouffard:
This PR seeks to replace Avro with ProtoBuf as the means to serialize/deserialize data. By switching to ProtoBuf, we should see an improved performance when sending things between Python and Scalal.

This PR resolves #199 #192 #191


jbouffard:
@echeipesh Since there is a lot in this PR, would you like to review it together? It might make the process a bit faster and easier.

jbouffard:
@fosskers Could you please review the `prorto` files and the encoding/decoding logic of the ProtBuf files in the Scala code when you have the chance?

geopyspark/protobufregistry.py
```diff
@@ -0,0 +1,455 @@
...
+        """Encodes a ``SpatialKey`` into bytes.
+
+        Args:
+            obj (:class:`~geopyspark.geotrellis.SpatialKey`): An instance of ``SpatialKey``.
+
+        Returns:
+            bytes
+        """
+
+        return obj.to_protobuf_spatial_key.SerializeToString()
```
echeipesh: this method should just be inlined here and avoid making named tuples into classes.

geopyspark/protobufregistry.py
```diff
@@ -0,0 +1,455 @@
...
+                                   SpaceTimeKey)
+
+from geopyspark.protobuf.tileMessages_pb2 import ProtoTile, ProtoMultibandTile, ProtoCellType
+from geopyspark.protobuf.extentMessages_pb2 import (ProtoExtent, ProtoProjectedExtent,
+                                                    ProtoTemporalProjectedExtent)
+from geopyspark.protobuf import keyMessages_pb2
+from geopyspark.protobuf import tupleMessages_pb2
+
+
+class ProtoBufRegistry(object):
```
echeipesh: This can now become a module named `protobuf_codecs` with `_*_decoder` functions becoming `from_pb_tile` and reverse of `to_pb_tile`.

geopyspark/protobufregistry.py
```diff
@@ -0,0 +1,455 @@
...
+
+        if tile.cellType.hasNoData:
+            return {'data': arr, 'no_data_value': tile.cellType.nd, 'data_type': data_type}
+        else:
+            return {'data': arr, 'data_type': data_type}
+
+    @classmethod
+    def _multibandtile_decoder(cls, multibandtile):
+        data_type = cls._mapped_data_types[multibandtile.tiles[0].cellType.dataType]
+        tiles = np.array([cls._tile_decoder(tile, data_type) for tile in multibandtile.tiles])
```
echeipesh: `tiles` here is actually `bands`

geopyspark/protobufregistry.py
```diff
@@ -0,0 +1,455 @@
...
+        Returns:
+            :ref:`Tile <raster>`
+        """
+
+        tile = ProtoTile.FromString(proto_bytes)
+        data_type = cls._mapped_data_types[tile.cellType.dataType]
+        arr = np.array([cls._tile_decoder(tile, data_type)])
+
+        if tile.cellType.hasNoData:
+            return {'data': arr, 'no_data_value': tile.cellType.nd, 'data_type': data_type}
```
echeipesh: We should not have a namedtuple for `Tile` where `no_data_value` can be None.

geopyspark/geopycontext.py
```diff
@@ -85,34 +85,17 @@ def map_key_input(key_type, is_boundable):
...
-                self.avroregistry.create_partial_tuple_encoder(key_type=key_type,
-                                                               value_type=value_type)
-
-        return AutoBatchedSerializer(AvroSerializer(schema, decoder, encoder))
+        return AutoBatchedSerializer(ProtoBufSerializer(decoder, encoder))
 
-    def create_value_serializer(self, schema, value_type):
-        decoder = self.avroregistry._get_decoder(value_type)
-        encoder = self.avroregistry._get_encoder(value_type)
+    def create_value_serializer(self, value_type):
```
echeipesh: Now that it is no longer possible or needed to register your own serializer these three methods into `protobufserializer.py`

geopyspark/protobufregistry.py
echeipesh: This and the protobufserializer should be moved into `geotrellis` module because it contains only codecs for geotrellis types.

PR #246, opened 2017-06-14
JVM TMS server

jpolchlo:
Provides a TMS server implementation in Scala for use through a GeoNotebook-enabled GeoPySpark jupyter notebook.  The implementation provided here allows for custom python rendering functions (at the cost of performance), and a default Scala-based rendering method using ColorMaps.

 - [x] Color mapping interface

Moved color mapping features from `rdd.py` to `color.py`, and added an interface to both `matplotlib` and `colortools` for ease of use.  The latter is a "soft dependency", that does not add `colortools` to the requirements.txt for GeoPySpark.  It may also be beneficially to apply this approach to the `colour` package.

 - [X] Proof of concept implementation using S3 catalogs

Experiments show that rendering using python takes 1-2 __seconds__ per tile, almost entirely due to the transmission of data from Java to Python (a known problem in Py4J, hoping to be fixed in version 0.11).  Native Scala rendering with colormaps renders tiles in <100 ms.

 - [x] Updated GeoNotebook interface allowing access to external TMS servers

There is more to do subsequent to this—specifically, the ability to serve a pyramided RDD—but these further achievements will require enhancements to GeoPySpark proper.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ColorMap.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ColorMap.scala
@@ -7,14 +7,6 @@ import scala.collection.JavaConversions._
 
 object ColorMap {
 
-  def stringToBoundaryType(s: String) = s.toLowerCase match {
-    case "greaterthan" => render.GreaterThan
-    case "greaterthanorequalto" => render.GreaterThanOrEqualTo
-    case "lessthan" => render.LessThan
-    case "lessthanorequalto" => render.LessThanOrEqualTo
-    case "exact" => render.Exact
-  }
-
   def importInt(x: Any): Int = {
     if (x.isInstanceOf[Int])
       x.asInstanceOf[Int]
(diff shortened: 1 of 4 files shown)


jpolchlo:
This works now.  Needs a TMS route instance that can serve from a Scala RDD.  That will require some coding effort and should most likely be done in a separate PR.  I'll take the WIP tag off this PR until someone says to put it back.

jpolchlo:
These failures are because repo.locationtech.com was down.  Travis should be kicked again when that resource is available again (it's still down at the time of this comment).

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ColorMap.scala
```diff
@@ -0,0 +1,51 @@
+package geopyspark.geotrellis
+
+import geotrellis.raster.render
+
+import scala.collection.JavaConverters._
+import scala.collection.JavaConversions._
+
+object ColorMap {
+
+  def stringToBoundaryType(s: String) = s.toLowerCase match {
```
jbouffard: We already have a method that does this https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/GeoTrellisUtils.scala#L83

geopyspark/geotrellis/tms.py
```diff
@@ -0,0 +1,56 @@
+from PIL import Image
+import numpy as np
+import io
+
+from py4j.clientserver import ClientServer, JavaParameters, PythonParameters
+
+# Return URL of tile server for a layer
+## http://localhost:2342/s3/azavea-datahub/catalog/tms/{z}/{x}/{y}
+def layer_tms(rdd_type, uri, layer_name, render_function=None, cmap=None):
```
jbouffard: What is the point of this function? It just `pass`es, and I don't see it being used anywhere where it's being implemented.

geopyspark/geotrellis/tms.py
```diff
@@ -0,0 +1,56 @@
...
+    # we neeed the setup so we can report common errors
+    pass
+
+def use_case():
+    url = layer_tms(SPATIAL, "s3://azavea-datahub/catalog", "us-ned-tms-epsg3857", cmap="viridean")
+
+class TileRender(object):
```
jbouffard: Docstrings for this class and its methods are needed.

geopyspark/geopycontext.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, pysc=None, **kwargs):
         self.sc = self.pysc._jsc.sc()
         self._jvm = self.pysc._gateway.jvm
 
+        self.pysc._gateway.start_callback_server()
```
jbouffard: Do you think it would be better to start the callback server only when someone is trying to a tms server?

geopyspark/geotrellis/color.py
```diff
@@ -0,0 +1,143 @@
...
+    Returns:
+        [str]
+    """
+
+    if num_colors:
+        return list(geopysc._jvm.geopyspark.geotrellis.ColorRamp.getHex(ramp_name, num_colors))
+    else:
+        return list(geopysc._jvm.geopyspark.geotrellis.ColorRamp.getHex(ramp_name))
+
+class ColorMap(object):
```
jbouffard: Docstrings for this class and its methods are needed.

geopyspark/geotrellis/color.py
```diff
@@ -0,0 +1,143 @@
...
+    @classmethod
+    def from_colors(cls, geopysc, breaks, color_list, noDataColor=0x00000000, fallback=0x00000000, class_boundary_type=LESSTHANOREQUALTO):
+        if all(isinstance(x, int) for x in breaks):
+            return cls(geopysc._jvm.geopyspark.geotrellis.ColorMap.fromBreaks(breaks, color_list, noDataColor, fallback, class_boundary_type))
+        else:
+            return cls(geopysc._jvm.geopyspark.geotrellis.ColorMap.fromBreaksDouble([float(br) for br in breaks], color_list, noDataColor, fallback, class_boundary_type))
+
+    @classmethod
+    def from_histogram(cls, geopysc, histogram, color_list, noDataColor=0x00000000, fallback=0x00000000, class_boundary_type=LESSTHANOREQUALTO):
+        pass
```
jbouffard: I'd leave some note here about how this hasn't been implemented yet.

geopyspark/geotrellis/color.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @staticmethod
+    def nlcd_colormap(geopysc):
+        nlcd_color_map =  { 0  : 0x00000000,
```
jbouffard: This might be better defined elsewhere so that `nlcd_color_map` isn't being defined each time someone calls `nlcd_colormap`.

geopyspark/geopycontext.py
jpolchlo: Could be done, but there may be value in just starting this server as a default action.  Might make it easy in the future for later applications that need to use the callback mechanism?

Maybe @echeipesh can weigh in on this?

PR #252, opened 2017-06-19
start_zoom is Now Optional When Pyramiding

### LINKED ISSUE #251: Default start_zoom in reproject if TiledRDD has a zoom level
Example:

```
def ingest(args):
    geopysc = GeoPyContext(appName="spacenet-ingest")

    print(args.input)
    # Read the GeoTiff from S3
    rdd = get(geopysc, SPATIAL, args.input)
    # rdd = get(geopysc, SPATIAL, "/Users/rob/proj/az/spacenet-viz/data/paris.tif")

    metadata = rdd.collect_metadata()

    # tile the rdd to the layout defined in the metadata
    laid_out = rdd.tile_to_layout(metadata)

    # reproject the tiled rasters using a ZoomedLayoutScheme
    reprojected = laid_out.reproject("EPSG:3857", scheme=ZOOM)

    # pyramid the TiledRasterRDD to create 12 new TiledRasterRDDs
    # one for each zoom level
    pyramided = reprojected.pyramid(end_zoom=1)

```

Fails because I need to supply a start zoom; ideally if I don't supply the start zoom it should use the zoom level attached to the `reproject` layer.

jbouffard:
This PR makes it so that a use doesn't have to input a `start_zoom` when pyramiding if their particular instance of `TiledRasterRDD` already has a `zoom_level`.

This Pr resolves #251

### DIFF
--- a/geopyspark/geotrellis/rdd.py
+++ b/geopyspark/geotrellis/rdd.py
@@ -701,14 +701,15 @@ class TiledRasterRDD(CachableRDD):
 
         return TiledRasterRDD(self.geopysc, self.rdd_type, srdd)
 
-    def pyramid(self, start_zoom, end_zoom, resample_method=NEARESTNEIGHBOR):
+    def pyramid(self, end_zoom, start_zoom=None, resample_method=NEARESTNEIGHBOR):
         """Creates a pyramid of GeoTrellis layers where each layer reprsents a given zoom.
 
         Args:
-            start_zoom (int): The zoom level where pyramiding should begin. Represents
-                the level that is most zoomed in.
             end_zoom (int): The zoom level where pyramiding should end. Represents
                 the level that is most zoomed out.
+            start_zoom (int, Optional): The zoom level where pyramiding should begin. Represents
+                the level that is most zoomed in. If None, then will use the zoom level from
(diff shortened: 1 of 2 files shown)


geopyspark/geotrellis/rdd.py
```diff
@@ -731,6 +732,13 @@ def pyramid(self, start_zoom, end_zoom, resample_method=NEARESTNEIGHBOR):
         if (num_cols & (num_cols - 1)) != 0 or (num_rows & (num_rows - 1)) != 0:
             raise ValueError("Tiles must have a col and row count that is a power of 2")
 
+        if start_zoom:
+            start_zoom = start_zoom
```
echeipesh: Does this do something magic or is it assignment to self?

jbouffard: It's just an assignment to self. There's similar code in other parts of GeoPySpark.

echeipesh: I don't follow the pattern, why assign to self here?

jbouffard: There isn't a really compelling reason to do it that way, it's just one of two implementations of that check.

PR #255, opened 2017-06-20
Fixed Imports

jbouffard:
This PR makes it so that when a user does `from geopyspark import *` or `from geopyspark.geotrellis import *` the moduels will be imported instead of just the packages themselves. In addition, this provides some convenience as the user can now shorten module paths when importing.

What is was like before:
```python
from geopyspark.geopycontext import GeoPyContext
from geopyspark.protobufregistry import ProtoBufRegistry
from geopyspark.protobufserializer import ProtoBufSerializer
```

What the imports are like now:
```python
from geopyspark import GeoPyContext, ProtoBufRegistry, ProtoBufSerializer
```

Read more about this [here](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package). Note, this PR is based on a current, open PR #245, and should not be merged until that one is.


jbouffard:
Superseded by #260

PR #260, opened 2017-06-20
Better Imports

jbouffard:
This PR supersedes #255 . It is now possible to import `ProtoBufSerializer`, `SPATIAL`, `SPACETIME`, `ZOOM`, `RasterRDD`, `TiledRasterRDD`, `PngRDD`, `Square`, `Circle`, `Nesw`, `Wedge`, and `Annulus` directly from `geopyspark.geotrellis`.

Before:
```python
from geopyspark.geotrellis.constants import SPATIAL, ZOOM, SLOPE
from geopyspark.geotrellis.neighborhoods import Square
from geopyspark.geotrellis.rdd import TiledRasterRDD
from geopyspark.geotrellis.render import PngRDD
```

Now:
```python
from geopyspark.geotrellis import SPATIAL, ZOOM, Square, TiledRasterRDD, PngRDD
from geopyspark.geotrellis.constants import SLOPE
```

This PR should be merged after #245

### DIFF
--- a/docker/notebooks/Chattanooga Ingest.ipynb
+++ b/docker/notebooks/Chattanooga Ingest.ipynb
@@ -10,10 +10,9 @@
    "source": [
     "from pyspark import SparkContext\n",
     "from geopyspark.geopycontext import GeoPyContext\n",
+    "from geopyspark.geotrellis import SPATIAL, ZOOM, RasterRDD, TiledRasterRDD\n",
     "from geopyspark.geotrellis.catalog import read, read_value, query, write\n",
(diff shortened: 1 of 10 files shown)


jbouffard:
@echeipesh This should be good to merge

jbouffard:
@echeipesh Right now, the only things whose imports have been made easier are classes and a few constants. Are there other constants that you'd like to see added? @jpolchlo believes the color ramp constants to be included as well.

And a question to both of you, should we make it so that functions like `get`, `read`, etc. can be imported via this method as well?

echeipesh:
Lets talk about imports after we figure out what is staying API and not mess around with it until then.

PR #261, opened 2017-06-20
Tiles are Now namedtuples

jbouffard:
This PR makes it so that `Tile`s are now represented by `namedtuple`s instead of `dict`s. While all of the code within the GeoPySpark code base works, any outside code will most likely break when this is merged.

This PR should be merged after #245


echeipesh:
@jpolchlo Tagging you for naming discussion.
The `Tile` type is needed, much more usable than a dictionary.
I'm thinking the `data` field should be called `cells` to make less generic and to change the `data_type` to `cell_type` to sync it with GT usage.

You can see the conventions clashing already here `return Tile(arr, tile.cellType.nd, None)`

Order  now matters too so I suggest to use`Tile(cells, cell_type, no_data_value)` with `no_data_value` defaulting to `None`.

jbouffard:
Those suggestions make sense to me. We're already breaking the API, so changing the field names isn't a really big deal at this point.

jpolchlo:
Yeah, I can get behind your proposal @echeipesh.  Seems logical/intuitive.

jbouffard:
@echeipesh @jpolchlo This has been updated to meet this PR's feedback. Ideally, I liked to merge this before merging #289. And once the aforementioned PR is merged, I'll redo #261

PR #263, opened 2017-06-21
GeoPyContext Removed

### LINKED ISSUE #258: Remove GeoPyContext
`GeoPyContext` is a remnant from an older implementation of GeoPySpark that should be removed. It should be replaced with a regular `SparkContext`. Any additional functionality it has can be moved to other modules. While this task is not difficult, it will be time consuming to update all of the code and docs.

jbouffard:
This Pr removes `GeoPyContext` and replaces it with a normal `SparkContext` instead. `GeoPyContext` is a remnant from an older implementation of GeoPyContext, and therefore, needs to be removed. **Once this is merged, code that uses GeoPySpark will break**. Luckily, resolving these issues should be straight forward.

If your code looked like this before:
```python
geopysc = GeoPyContext(appName="example", master="local[*]")

# or
pysc = SparkContext(appName="example", master="local[*]")
geopysc = GeoPyContext(pysc)
```

Then needs to be changed to:
```python
pysc = SparkContext(appName="example", master="local[*]")
```
With all references to `geopysc` being replaced with `pysc`.

This Pr resolves #258


jbouffard:
Superseded by #292

Issue #282, opened 2017-06-22
Break out non-mapalgebra functions into their own files


echeipesh:
```
geotrellis/
        __init__.py
        layer.py
                CachableLayer
                RasterLayer
                TiledRasterLayer
                Pyramid
        geotiff.py
        hillshade.py
        cost_distance.py
        euclidean_distance.py
        rasterize.py
```
Use the `__all__` in core module to keep imports tidy

jbouffard:
Resolved in #289

Issue #271, opened 2017-06-22
Prefer `neighborhood` package name


jpolchlo:
Currently, we have a `neighborhoods.py` file; this should be renamed to `neighborhood.py` to make imports more natural.  I.e., `neighborhood.Square` is more grammatically correct.

jbouffard:
Resolved in #289

PR #289, opened 2017-06-23
Renames and Breakouts

### LINKED ISSUE #283: Rename RasterRDD and friends to RasterLayer
The thinking is: its not an RDD, it doesn't even have a python RDD. It has a reference to scala RDD. But even now we can imagine it having a reference to ModelLab-like structure that can build an RDD remotely.

jbouffard:
This PR renames various modules and classes. The following modules have been renamed: `rdd` -> `layer`, `geotiff_rdd` -> `geotiff`, `neighborhoods` -> `neighborhood`. These are the classes that have had their names changed: `RasterRDD` -> `RasterLayer`, `TiledRasterRDD` -> `TiledRasterLayer`, `CachableRDD` -> `CachableLayer`.

In addition to renames, this PR also breaks out certain methods of `TiledRasterRDD` into their own modules. These new modules are: `cost_distance`, `euclidean_distance`, `hillshade`, and `rasterize`.

This Pr resolves #283 #282 #271


echeipesh:
This looks good to me. The workbooks are all going out of sync, but we should just fix them in one step after we finish with the rest of the changes I think.

echeipesh:
Also this is going to clash really hard with removing GeoPySpark PR : (

jbouffard:
@echeipesh Yeah, I was thinking we could update the notebooks and docs all in one step for each. That's very true :frowning_face: Rebasing #261 and #263 on top of the current master is already proving to be an annoying task to resolve.

PR #291, opened 2017-06-23
[WIP] Streamline inclusion of third-party jars

jpolchlo:
We need a simplified means of passing third-party Java code in .jar files to GeoPySpark scripts.  This PR will facilitate that by providing a well-defined pathway for that inclusion.

This PR also contributes to the move away from GeoPyContext to plain SparkContexts by providing the means to build basic SparkConfs that the user can manipulate (or not) before passing to the SparkContext constructor.


jamesmcclain:
The PySpark context object has an `addJar` method.  That might be worth investigation.

jamesmcclain:
Just to follow up on our discussion: If the targets that produce the jar file are left in the makefile, that is good enough for me.  Possibly more convenient would be to include the jar in the image, but that is not a requirement for me.

geopyspark/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+    if jar_dir not in local_prefixes:
+        possible_jars.append(jar)
+
+    returned = [glob.glob(jar_files) for jar_files in possible_jars]
+    jars = [jar for sublist in returned for jar in sublist]
+
+    if len(jars) == 0:
+        raise IOError("Failed to find any jars. Looked at these paths {}".format(possible_jars))
+
+    jar_string = str(jars[0])
```
jbouffard: This will need to be changed if we want to support multiple jars. The way it is now is that it'll just take the first jar of the list regardless of how many were actually found.

I believe something like this should work:

```python
jar_string = ", ".jon(jars)
```

jamesmcclain: `join`?

jbouffard: Yeah, I meant `join` and not `jon`.

geopyspark/geopycontext.py
```diff
@@ -0,0 +1,106 @@
...
+import geopyspark.geotrellis.converters
+
+ensure_pyspark()
+
+from pyspark import RDD, SparkContext
+from pyspark.serializers import AutoBatchedSerializer
+
+from py4j.java_gateway import java_import
+
+class GeoPyContext(object):
```
jbouffard: Does `GeoPyContext` need to be added back in?

geopyspark/geopyspark_utils.py
```diff
@@ -9,15 +9,18 @@
 from geopyspark.geopyspark_constants import JAR
 
 
-def check_environment():
-    jars = 'JARS' in os.environ
-    pyspark_args = 'PYSPARK_SUBMIT_ARGS' in os.environ # driver (YARN)
-    yarn = ('SPARK_YARN_MODE' in os.environ) and \
-            (os.environ['SPARK_YARN_MODE'] == 'true') # executor (YARN)
+# def check_environment():
```
jbouffard: If we don't need this then could you please delete it?

geopyspark/geopyspark_utils.py
```diff
@@ -43,59 +46,61 @@ def add_pyspark_path():
         raise ValueError("Could not find the py4j zip in", path.join(pyspark_home, 'python', 'lib'))
 
 
-def setup_environment():
-    """Sets up various environment variables that are needed to run GeoPySpark.
+# def setup_environment():
```
jbouffard: Same as above comment.

geopyspark/tests/__init__.py
```diff
@@ -1,6 +1,6 @@
 import os
 
 
-if 'JARS' not in os.environ:
-    from geopyspark.geopyspark_utils import setup_environment
-    setup_environment()
+# if 'JARS' not in os.environ:
```
jbouffard: Same as the above two comments as well.

geopyspark/tests/base_test_class.py
```diff
@@ -18,7 +19,8 @@ class BaseTestClass(unittest.TestCase):
     else:
         master_str = "local[*]"
 
-    pysc = SparkContext(master=master_str, appName="test")
+    conf = initial_spark_conf(master=master_str, appName="test")
```
echeipesh: Lets name it `geopyspark_conf`

Issue #272, opened 2017-06-22
Remove TILE constant


jpolchlo:
Not useful anymore.  Remove this from `constants.py`.

jbouffard:
Resolved in #293

PR #293, opened 2017-06-26
Constants Orginization

### LINKED ISSUE #274: Constant grouping for ipython tab completion
Certain functions have parameters that take certain named constant arguments.  Currently, all of our constants are lumped together in `constants.py`.  This would suggest that a user might not always know which constant to use at a quick glance for a specific purpose.  To avoid confusion, we should figure out how to better group constants into semantic units and enable tab-completion on those groups in iPython.

jbouffard:
This PR organizes most constants so that they are grouped by a class. For instance, instead of calling `from geopyspark.geotrellis.constants import NEARESTNEIGHBHOR`, it will now be `from geopyspark.geotrellis.constants import ResampleMethods` which the user can then do `ResampleMethods.NEARESTNEIGHBHOR` to specify which method to use. This should make it easier to find/discover the various constants available in GeoPySpark.

In addition to these new groupings, the `TILE` constant has also been removed.

This PR resolves #274 #272

### DIFF
--- a/geopyspark/geotrellis/constants.py
+++ b/geopyspark/geotrellis/constants.py
@@ -60,18 +60,17 @@ class ResampleMethods(object):
     MAX = 'Max'
     MIN = 'Min'
 
-
-RESAMPLE_METHODS = [
-    ResampleMethods.NEARESTNEIGHBOR,
-    ResampleMethods.BILINEAR,
-    ResampleMethods.CUBICCONVOLUTION,
-    ResampleMethods.LANCZOS,
-    ResampleMethods.AVERAGE,
-    ResampleMethods.MODE,
-    ResampleMethods.MEDIAN,
-    ResampleMethods.MAX,
-    ResampleMethods.MIN
-]
+    RESAMPLE_METHODS = [
+        NEARESTNEIGHBOR,
+        BILINEAR,
+        CUBICCONVOLUTION,
+        LANCZOS,
+        AVERAGE,
+        MODE,
+        MEDIAN,
+        MAX,
+        MIN
+    ]
 
 
 class TimeUnits(object):
@@ -85,19 +84,18 @@ class TimeUnits(object):
     MONTHS = 'months'
     YEARS = 'years'
 
+    TIME_UNITS = [
+        MILLISECONDS,
+        SECONDS,
+        MINUTES,
+        HOURS,
+        DAYS,
+        MONTHS,
+        YEARS
+    ]
 
-TIME_UNITS = [
-    TimeUnits.MILLISECONDS,
-    TimeUnits.SECONDS,
-    TimeUnits.MINUTES,
-    TimeUnits.HOURS,
-    TimeUnits.DAYS,
-    TimeUnits.MONTHS,
-    TimeUnits.YEARS
-]
 
-
-class Operations(object):
+class Operation(object):
     """Focal opertions."""
 
     SUM = 'Sum'
@@ -110,21 +108,20 @@ class Operations(object):
     SLOPE = 'Slope'
     STANDARDDEVIATION = 'StandardDeviation'
 
+    OPERATIONS = [
+        SUM,
+        MIN,
+        MAX,
+        MEAN,
+        MEDIAN,
+        MODE,
+        STANDARDDEVIATION,
+        ASPECT,
+        SLOPE
+    ]
 
-OPERATIONS = [
-    Operations.SUM,
-    Operations.MIN,
-    Operations.MAX,
-    Operations.MEAN,
-    Operations.MEDIAN,
-    Operations.MODE,
-    Operations.STANDARDDEVIATION,
-    Operations.ASPECT,
-    Operations.SLOPE
-]
 
-
-class Neighborhoods(object):
+class Neighborhood(object):
     """Neighborhood types."""
 
     ANNULUS = 'annulus'
@@ -133,14 +130,13 @@ class Neighborhoods(object):
     WEDGE = 'wedge'
     CIRCLE = "circle"
 
-
-NEIGHBORHOODS = [
-    Neighborhoods.ANNULUS,
-    Neighborhoods.NESW,
-    Neighborhoods.SQUARE,
-    Neighborhoods.WEDGE,
-    Neighborhoods.CIRCLE
-]
+    NEIGHBORHOODS = [
+        ANNULUS,
+        NESW,
+        SQUARE,
+        WEDGE,
+        CIRCLE
+    ]
 
 
 class ClassificationStrategies(object):
@@ -152,14 +148,13 @@ class ClassificationStrategies(object):
     LESSTHANOREQUALTO = "LessThanOrEqualTo"
     EXACT = "Exact"
 
-
-CLASSIFCATION_STRATEGIES = [
-    ClassificationStrategies.GREATERTHAN,
-    ClassificationStrategies.GREATERTHANOREQUALTO,
-    ClassificationStrategies.LESSTHAN,
-    ClassificationStrategies.LESSTHANOREQUALTO,
-    ClassificationStrategies.EXACT
-]
+    CLASSIFCATION_STRATEGIES = [
+        GREATERTHAN,
+        GREATERTHANOREQUALTO,
+        LESSTHAN,
+        LESSTHANOREQUALTO,
+        EXACT
+    ]
 
 
 class CellTypes(object):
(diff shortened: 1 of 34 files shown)


geopyspark/geotrellis/constants.py
```diff
@@ -93,237 +42,207 @@
...
+    AVERAGE = 'Average'
+    MODE = 'Mode'
+    MEDIAN = 'Median'
+    MAX = 'Max'
+    MIN = 'Min'
 
 
-"""Neighborhood type."""
-ANNULUS = 'annulus'
+RESAMPLE_METHODS = [
```
echeipesh: Can these be moved into the class somehow so they can be references with `if x in ResampleMethods: ...`  ?

geopyspark/geotrellis/constants.py
```diff
@@ -93,237 +42,207 @@
...
-"""A classification strategy."""
-LESSTHANOREQUALTO = "LessThanOrEqualTo"
-
-"""A classification strategy."""
-EXACT = "Exact"
-
-
-"""Representes Bit Cells."""
-BOOLRAW = "boolraw"
+class Neighborhoods(object):
```
echeipesh: `Neighborhood` just like module name

geopyspark/geotrellis/constants.py
```diff
@@ -93,237 +42,207 @@
...
-    SUM,
-    MIN,
-    MAX,
-    MEAN,
-    MEDIAN,
-    MODE,
-    STANDARDDEVIATION,
-    ASPECT,
-    SLOPE
+    Operations.SUM,
```
echeipesh: This should be singular: `Operation` as well.

echeipesh:
Also noticed that we're not using (Neighborhood)[https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/neighborhood.py] class in the focal operations. Doesn't have to be part of this PR but should be addressed.

geopyspark/geotrellis/constants.py
jbouffard: Yeah, we can do that. I'll make the change now.

echeipesh:
Hmm, if we require 3.4 we get the nice `Enum` from standard library: https://www.python.org/dev/peps/pep-0435/

echeipesh:
Also we should make all enum values singular, `CellType`, `ResampleMethods` and the rest.

jbouffard:
@echeipesh One good thing about people refusing to upgrade their Python version is that many new features are back-ported, so we can use versions earlier than 3.4 and still use `Enum`.

jbouffard:
@echeipesh @jpolchlo One thing I'd like the user to be able to do is use either a string or its constant representation when calling methods. However, the way the strings are formatted for each group of constants is different. For example, see  [ReampleMethod](https://github.com/jbouffard/geopyspark/blob/4fd04b6b4073c3f78a82cd478a0d2e992826beee/geopyspark/geotrellis/constants.py#L54) and [TimeUnit](https://github.com/jbouffard/geopyspark/blob/4fd04b6b4073c3f78a82cd478a0d2e992826beee/geopyspark/geotrellis/constants.py#L81).

Do you guys think it'd be worth formatting these strings so that they're all the same, and if so, how should they be formatted?

echeipesh:
Keeping the same convention for constants is a good thing, we should do it as much as possible. I would even be tempted to keep the capitalization on the variables to be the same as string constants ex: `ResampleMethod.NearestNeighbor`.

But `CellType` for instance should be the way it is, because it matches convention used by GT tiles.

jbouffard:
I agree with changing the capitalization of the variables to match the strings. I'll make that update now.

jbouffard:
@echeipesh This should be good for review now.

geopyspark/geotrellis/constants.py
```diff
@@ -93,237 +44,200 @@
...
+    CubicConvolution = 'CubicConvolution'
+    CubicSpline = 'CubicSpline'
+    Lanczos = 'Lanczos'
+    Average = 'Average'
+    Mode = 'Mode'
+    Median = 'Median'
+    Max = 'Max'
+    Min = 'Min'
+
+    RESAMPLE_METHODS = [
```
CloudNiner: Unless it's necessary for backwards compatible reasons, I don't think it's necessary to keep these list declarations. I see further down that you do a `if resample_method not in ResampleMethod.RESAMPLE_METHODS.values:`. If `resample_method` is an instance of one of the concrete options of `ResampleMethod` then you can just do `if resample_method not in ResampleMethod`. Same thing anywhere you need to iterate over all of the options of a given enum. The class itself implements iterator, so you can do:
```
for r in ResampleMethod:
    print(r)
```
and you'll iterate the list of options.

If you do need to keep the list for some reason, consider replacing each entry with the class variable rather than the string, e.g.:
```
RESAMPLE_METHODS = [
    NearestNeighbor,
    ...,
]
```

geopyspark/geotrellis/constants.py
```diff
@@ -93,237 +44,200 @@
...
-    LIGHT_TO_DARK_SUNSET,
-    LIGHT_TO_DARK_GREEN,
-    HEATMAP_YELLOW_TO_RED,
-    HEATMAP_BLUE_TO_YELLOW_TO_RED_SPECTRUM,
-    HEATMAP_DARK_RED_TO_YELLOW_WHITE,
-    HEATMAP_LIGHT_PURPLE_TO_DARK_PURPLE_TO_WHITE,
-    CLASSIFICATION_BOLD_LAND_USE,
-    CLASSIFICATION_MUTED_TERRAIN
-]
+class ResampleMethod(Enum):
```
CloudNiner: Use of Enum ok if you're planning on restricting use of this code to Python 3.4+. @echeipesh already mentioned that you're only supporting Python 3, so this seems ok given that. Python 3.4 is no longer supported, and 3.5/3.6 are fully backwards compatible. Just wanted to be sure this was noted.

Also, now that you're using `enum`, the `setup.py` file should be updated to indicate that this package requires Python 3.4+.

geopyspark/geotrellis/constants.py
```diff
@@ -93,237 +44,200 @@
...
+    """Resampling Methods."""
+
+    NearestNeighbor = 'NearestNeighbor'
```
CloudNiner: Constants in Python are typically upper case, words separated by underscores. Consider updating.

See: https://www.python.org/dev/peps/pep-0008/#constants

geopyspark/tests/png_test.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 
 from geopyspark.geotrellis import Extent, ProjectedExtent, Tile
-from geopyspark.geotrellis.constants import SPATIAL, HOT
+#from geopyspark.geotrellis.constants import SPATIAL, HOT
```
CloudNiner: Can this be removed rather than commented out?

geopyspark/geotrellis/constants.py
jbouffard: I believe that `Enum` has been ported to early versions of Python. For example:

```
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from enum import Enum
>>> class Test(Enum):
...     Red = "Red"
... 
>>> Test.Red
<Test.Red: 'Red'>
>>> type(Test.Red)
<enum 'Test'>
```

Unless `Enum` in Python 3.3 is less feature complete than Python 3.4+, then I think the `setup.py` should be okay to stay how it is.

geopyspark/geotrellis/constants.py
jbouffard: Originally, constants were all caps. However, we wanted the user to be able to enter a string in lieu of a constant value, so we thought it'd be best to have the formatting of the constants match their respective string.

geopyspark/geotrellis/constants.py
jbouffard: As mentioned above, we'd like to keep the option of letting the user input a string instead of constant if they so desire. That's the main reason for keeping those lists around, as `"NearestNeighbor" in ResampleMethod` will always return `False`.

geopyspark/tests/png_test.py
jbouffard: Yeah, this can be removed. However, this entire file is actually deleted in another PR #296, so I just commented it out in this one just so the tests can pass.

geopyspark/geotrellis/constants.py
CloudNiner: Gotcha. I haven't used this feature in earlier versions of python 3 so I don't have experience with that. It looks like your test runner has a series of tests that use these new constants, so if those pass on earlier versions of Python 3 it seems like it would be ok to leave version support as is.

geopyspark/geotrellis/constants.py
CloudNiner: Ah, I see. That works. 

For completeness sake, one available alternative would be:
```
class ResampleMethod(Enum):
    NEAREST_NEIGHBOR = 'NearestNeighbor'
    ...
```
That way you could still instantiate enum instances with `ResampleMethod('NearestNeighbor')` but maintain the existing python constant convention.

This also dovetails nicely into my comment below about removing the RESAMPLE_METHODS list, as it looks like its just used to determine if a string is in the enum elsewhere in your code. That can be done with the same `ResampleMethod('NearestNeighbor')` as described above, and helpfully throws a `ValueError: {value} is not a valid {enumclass}`

jbouffard: I think that's a pretty good solution. The only issue I can see is that the value being sent over to Scala needs to be a string, so you'd have to do: `str -> ResampleMethod -> str` if the user gave a string as the argument. It isn't that much overhead, but it would be nice to great rid of the middle step.

What are your thoughts on this @echeipesh?

jbouffard:
@echeipesh @CloudNiner After thinking about it, I realized there really is no point in making the constant classes subclasses of `Enum`. They should just be regular classes instead.

`Enum` classes are somewhat awkward to use in GeoPySpark. For example, `ResampleMethods.NearestNeighbor` returns a `ResampleMethods` type and not a `str`. This is a bit annoying since the type of the value that needs to be sent over has to be a `str`. So that value needs to be converted to a `str`, which is easy enough, but I see no reason to have this conversion step. Even if the user were to input a `str` in place of the constant, the values within the constant class would still need to be converted to a collection of `str` in order to do the check (ie. `my_string in ResampleMethod.RESAMPLE_METHODS.values`).

Using normal classes would be more straight forward:

```python3
ResampleMethod(object):
    NEAREST_NEIGHBOR = "NearestNeighbor"
    # rest of the methods
```

Now, `ResampleMethod.NEAREST_NEIGHBOR` is a `str`, which makes it easier to work with.

What are your guys' thoughts?

PR #294, opened 2017-06-26
Deprecated catalog.read in Favor of catalog.query

### LINKED ISSUE #280: Deprecate catalog.read in favor of catalog.query
The catalog read method in GT is not extremely useful for non-toy problems.  We should favor using query.  The function should also have its arguments renamed for clarity.  A suggested naming strategy would be
```python
def catalog.query(geopysc,
    uri,
    layer_name,
    layer_zoom,
    query_geom=None,
    time_intervals=None,
    query_proj=None,
    options=None,
    numPartitions=None,
    **kwargs)
```

jbouffard:
This PR deprecates `catalog.read` in favor of `catalog.query` instead. This simplifies the API as both functions do similar things.

This PR resolves #280

### DIFF
--- a/geopyspark/geotrellis/catalog.py
+++ b/geopyspark/geotrellis/catalog.py
@@ -309,9 +309,9 @@ def query(pysc,
           uri,
           layer_name,
           layer_zoom,
-          intersects=None,
+          query_geom=None,
           time_intervals=None,
-          proj_query=None,
+          query_proj=None,
           options=None,
           numPartitions=None,
           **kwargs):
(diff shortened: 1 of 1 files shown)


echeipesh:
We should go ahead with the renames for the parameters here too, `proj_query` is too confusing in retrospect.
```py
    query_geom=None,
    time_intervals=None,
    query_proj=None,
    options=None,
    numPartitions=None,
``

jbouffard:
@echeipesh I forgot about that. My bad. Just updated the PR so it should be good to go.

PR #301, opened 2017-06-27
Histogram Wrapper

### LINKED ISSUE #286: Python representation for Histogram
Right now `.get_histogram` gives a `JavaObject` we get from py4j call. 
We need to have a wrapper around it that exposes some functionality about it and a way to construct it.

jbouffard:
This PR adds the `Histogram` class to GeoPySpark. This acts as a wrapper for GeoTrellis' `Histogram[T]` type. This class can do many of the same tasks its Scala counterpart can do, and is what's expected and returned when dealing with histograms in GeoPySpark.

This PR resolves #286

### DIFF
--- a/geopyspark/geotrellis/histogram.py
+++ b/geopyspark/geotrellis/histogram.py
@@ -21,20 +21,6 @@ class Histogram(object):
     def __init__(self, scala_histogram):
         self.scala_histogram = scala_histogram
 
-    @classmethod
-    def from_tiled_layer(cls, tiled_layer):
-        """Creates an instance of ``Histogram`` from a ``TiledRasterLayer``.
-
-        Args:
-            tiled_layer (:class:`~geopyspark.geotrellis.layer.TiledRasterLayer`): The tiled layer
-                from which the histogram should be derived from.
-
-        Returns:
-            :class:`~geopyspark.geotrellis.histogram.Histogram`
-        """
-
-        return cls(tiled_layer.get_histogram())
-
     def min(self):
         """The smallest value of the histogram.
 
--- a/geopyspark/geotrellis/histogram.py
+++ b/geopyspark/geotrellis/histogram.py
@@ -1,0 +1,46 @@
+
+
+class Histogram(object):
+    def __init__(self, scala_histogram):
+        self.scala_histogram = scala_histogram
+
+    @classmethod
+    def from_tiled_layer(cls, tiled_layer):
+        return cls(tiled_layer.get_histogram())
+
+    def min(self):
+        return self.scala_histogram.minValue().get()
+
+    def max(self):
+        return self.scala_histogram.maxValue().get()
+
+    def min_max(self):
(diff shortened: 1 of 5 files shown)


geopyspark/geotrellis/histogram.py
```diff
@@ -0,0 +1,159 @@
...
+    def bucket_count(self):
+        """Returns the number of buckets within the histogram.
+
+        Returns:
+            int
+        """
+
+        return self.scala_histogram.bucketCount()
+
+    def quantile_breaks(self):
```
jpolchlo: This function appears to want an argument—`num_breaks` as listed in the doc string—but none is given or passed to the underlying object.

geopyspark/geotrellis/layer.py
```diff
@@ -1041,13 +1042,13 @@ def get_histogram(self):
             None
```
jpolchlo: Line 1039 was not adjusted.  Still says it returns an array of java objects.

geopyspark/geotrellis/histogram.py
```diff
@@ -135,11 +135,14 @@ def bucket_count(self):
...
         """Returns quantile breaks for this Layer.
 
         Args:
             num_breaks (int): The number of breaks to return.
+
+        Returns:
+            ``[int]``
         """
 
         return list(self.scala_histogram.quantileBreaks())
```
jpolchlo: Should `num_breaks` be passed to `self.scala_histogram.quantileBreaks`?

jbouffard: Yeah, it should be... my bad.

geopyspark/geotrellis/histogram.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Attributes:
+        scala_histogram (py4j.JavaObject): An instance of the GeoTrellis histogram.
+    """
+
+    def __init__(self, scala_histogram):
+        self.scala_histogram = scala_histogram
+
+    @classmethod
+    def from_tiled_layer(cls, tiled_layer):
```
echeipesh: Although you can't do it right now you should be able to get a histogram from `RasterLayer`. I would argue that we don't want to proliferate `from_*` for geopyspark classes that have `.get_histogram` methods.

jbouffard: That makes sense. I only put that there as an other alternative to constructing the `Histogram`. I'll remove it when I'm done with the constants PR.

echeipesh: right, the actual feedback is also that `TiledRasterLayer.get_histogram` should return this class.

jbouffard: It does already https://github.com/jbouffard/geopyspark/blob/d51b509ac533ed367d39abcabeb88f6ff4a273c9/geopyspark/geotrellis/layer.py#L1050

jbouffard:
@echeipesh This has been updated and rebased on top of the current master

PR #304, opened 2017-06-27
Clean up SparkConf construction & simplify third-party JAR usage

jpolchlo:
Following the move to explicit construction of SparkContext objects, we want to simplify the process of building SparkConf objects for use in GeoPySpark.  To that end, this PR provides the means to construct a base SparkConf that can then be manipulated by the user to suit their needs.  The lines
```python
from pyspark import SparkContext
from geopyspark import geopyspark_conf
conf = geopyspark_conf(appName='name', master='local[*]')
sc = SparkContext(conf=conf)
```
will provide a basic SparkContext.

This change also provides a convenient opening to update the way we handle third-party Java code.  We offer that JARs may be placed in the `jars` subdirectory off the current working directory (at the time of creating the SparkConf), or in any directory specified by the colon-separated `GEOPYSPARK_JARS_PATH` environment variable, and these will be placed in the SparkConf generated by `geopyspark_conf`.

It is important that any configs specified by the `PYSPARK_SUBMIT_ARGS` environment variable not specify `--jars`, or that setting will take precedence and require the manual addition of any extra jars.


docker/kernels/local/kernel.json
```diff
@@ -10,8 +10,9 @@
       "LD_LIBRARY_PATH": "/home/hadoop/local/gdal/lib",
       "PYSPARK_PYTHON": "/usr/bin/python3.4",
       "SPARK_HOME": "/usr/local/spark-2.1.0-bin-hadoop2.7",
+      "GEOPYSPARK_JARS_PATH": "/opt/jars",
       "PYTHONPATH": "/usr/local/spark-2.1.0-bin-hadoop2.7/python/lib/pyspark.zip:/usr/local/spark-2.1.0-bin-hadoop2.7/python/lib/py4j-0.10.4-src.zip",
-      "PYSPARK_SUBMIT_ARGS": "--master local[*] --archives /blobs/gdal-and-friends.tar.gz,/blobs/geopyspark-and-friends.tar.gz --jars /blobs/geotrellis-backend-assembly-0.1.0.jar --conf spark.serializer=org.apache.spark.serializer.KryoSerializer --conf spark.kyro.registrator=geotrellis.spark.io.kyro.KryoRegistrator --conf spark.executorEnv.LD_LIBRARY_PATH=/home/hadoop/local/gdal/lib/ pyspark-shell"
+      "PYSPARK_SUBMIT_ARGS": "--master local[*] --conf spark.executorEnv.LD_LIBRARY_PATH=/home/hadoop/local/gdal/lib/ pyspark-shell"
```
echeipesh: Is there still a point of specifying `--master` here since its going to happen in `geopyspark_conf` ?

echeipesh: Actually is the whole `PYSPARK_SUBMIT_ARGS` needed? `LD_LIBRARY_PATH` is given above as env var and I think `pyspark-shell` is assumed.

geopyspark/__init__.py
```diff
@@ -48,3 +49,78 @@ def create_python_rdd(pysc, jrdd, serializer):
         return RDD(jrdd, pysc, serializer)
     else:
         return RDD(jrdd, pysc, AutoBatchedSerializer(serializer))
+
+def geopyspark_conf(appName=None, master=None, additional_jar_dirs=[]):
```
echeipesh: `master` and `appName` should be reversed in order to match the order of `SparkContext` and minimize errors from muscle memory.

docker/kernels/local/kernel.json
jpolchlo: Yes.  The environment should specify the default value so that notebooks do not need to be altered depending on where it is run.

echeipesh: Ah, so the `SparkConf` the `SparkConf` will overwrite the mater. What about `--conf spark.executorEnv.LD_LIBRARY_PATH` is that dupe of `LD_LIBRARY_PATH` or is it needed for other reason?

echeipesh:
Verified that removing `--master` from `geonotebook-yarn/kernel.json` makes it usable in local case.

echeipesh:
Was able to use the same container on EMR, looks like we can get rid of `-local`/'-yarn` distinction.

docker/Makefile
```diff
@@ -7,7 +7,7 @@ NETCDF-JAR := netcdfAll-5.0.0-SNAPSHOT.jar
 GDAL-BLOB := archives/gdal-and-friends.tar.gz
 PYTHON-BLOB := archives/geopyspark-and-friends.tar.gz
 GEONOTEBOOK := 178bde1679ae3bc084c044ca3eeb915a137f689f
```
echeipesh: The `GEONOTEBOOK` SHA should be updated to `eeb1ee9e2dfb9df84eec24306c4b05b7008865ec` which is the merge of renamed changes: https://github.com/geotrellis/geonotebook/commit/eeb1ee9e2dfb9df84eec24306c4b05b7008865ec

PR #307, opened 2017-06-29
Remove Docker Image Build

jamesmcclain:
The docker image build has been moved [here](http://github.com/geodocker/geodocker-jupyter-geopyspark).


jpolchlo:
Working on bringing this up to date with current master.  Notebook runs, but something odd is happening in display, with significant computation at the point of serving each tile.  Possibly related to the calculation of histograms?  I'm working on it.

docker/kernels/geonotebook/kernel.json
```diff
@@ -1,20 +0,0 @@
-{
```
echeipesh: Pointing out an inconsistency here:
In the master we've been able to use this `kernel.json` on both local and yarn case
where is the `geodocker-jupyter-geonotebook` repo still has two kernels
https://github.com/geodocker/geodocker-jupyter-geopyspark/tree/master/kernels

jamesmcclain: Thank you, I have a branch that corrects that

jamesmcclain: https://github.com/geodocker/geodocker-jupyter-geopyspark/blob/master/kernels/geonotebook/kernel.json

jamesmcclain:
I think this is ready to be merged.

PR #313, opened 2017-06-30
to_png_rdd Method

### LINKED ISSUE #277: Add .to_png_rdd(cm) to RasterRDD and TiledRasterRDD
This should use ColorMap to return a python RDD of `(SpatialKey, bytes)` or `(SpaceTimeKey, bytes)` of the PNG. This will be using the Scala PNG renderer to make the export consistent with TMS view and be super fast.

jbouffard:
This PR adds the `to_png_rdd` method to `RasterLayer` and `TiledRasterLayer`. When called, this will return a `RDD[(K, bytes)]` where the `bytes` the PNG bytes of the raster and the `K` is the corresponding key of the PNG.

This PR resolves #277

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
@@ -11,6 +11,8 @@ import geotrellis.raster._
 import geotrellis.vector._
 import geotrellis.spark._
 
+import com.google.protobuf.ByteString
+
 
 trait TupleProtoBuf {
 
@@ -26,6 +28,17 @@ trait TupleProtoBuf {
         multibandTileProtoBufCodec.decode(message.tiles.get))
   }
 
+  implicit def tupleProjectedExtentImageProtoBufCodec =
+    new ProtoBufCodec[(ProjectedExtent, Array[Byte]), ProtoTuple] {
+    def encode(tuple: (ProjectedExtent, Array[Byte])): ProtoTuple =
+      ProtoTuple(
+        projectedExtent = Some(projectedExtentProtoBufCodec.encode(tuple._1)),
+        imageBytes = ByteString.copyFrom(tuple._2))
+
+    def decode(message: ProtoTuple): (ProjectedExtent, Array[Byte]) =
+      throw new Exception("This ProtoBuf message cannot be decoded")
+  }
+
   implicit def tupleTemporalProjectedExtentProtoBufCodec =
     new ProtoBufCodec[(TemporalProjectedExtent, MultibandTile), ProtoTuple] {
     def encode(tuple: (TemporalProjectedExtent, MultibandTile)): ProtoTuple =
@@ -36,6 +49,17 @@ trait TupleProtoBuf {
     def decode(message: ProtoTuple): (TemporalProjectedExtent, MultibandTile) =
       (temporalProjectedExtentProtoBufCodec.decode(message.temporalProjectedExtent.get),
         multibandTileProtoBufCodec.decode(message.tiles.get))
+  }
+
+  implicit def tupleTemporalProjectedExtentImageProtoBufCodec =
+    new ProtoBufCodec[(TemporalProjectedExtent, Array[Byte]), ProtoTuple] {
+    def encode(tuple: (TemporalProjectedExtent, Array[Byte])): ProtoTuple =
+      ProtoTuple(
+        temporalProjectedExtent = Some(temporalProjectedExtentProtoBufCodec.encode(tuple._1)),
+        imageBytes = ByteString.copyFrom(tuple._2))
+
+    def decode(message: ProtoTuple): (TemporalProjectedExtent, Array[Byte]) =
+      throw new Exception("This ProtoBuf message cannot be decoded")
   }
 
   implicit def tupleSpatialKeyProtoBufCodec =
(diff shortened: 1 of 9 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
```diff
@@ -50,6 +74,17 @@ trait TupleProtoBuf {
         multibandTileProtoBufCodec.decode(message.tiles.get))
   }
 
+  implicit def tupleSpatialKeyImageProtoBufCodec =
+    new ProtoBufCodec[(SpatialKey, Array[Byte]), ProtoTuple] {
```
echeipesh: You could save having to enumerate all the cases here by parameterizing the `K` to get signature like:

```
implicit def tupleSpatialKeyImageProtoBufCodec[K](implicit keyCodec: ProtoBufCodec[K]) =
  new ProtoBufCodec[(K, Array[Byte]), ProtoTuple] {
    def encode(tuple: (SpatialKey, Array[Byte])): ProtoTuple =
      ProtoTuple(
        spatialKey = Some(keyCodec.encode(tuple._1)),
        imageBytes = ByteString.copyFrom(tuple._2))

    def decode(message: ProtoTuple): (SpatialKey, Array[Byte]) = {
      ???
    }
  }  ...
```

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
```diff
@@ -50,6 +74,17 @@ trait TupleProtoBuf {
...
+    def encode(tuple: (SpatialKey, Array[Byte])): ProtoTuple =
+      ProtoTuple(
+        spatialKey = Some(spatialKeyProtoBufCodec.encode(tuple._1)),
+        imageBytes = ByteString.copyFrom(tuple._2))
```
echeipesh: `imageBytes` isn't the best name here, its already a stretch between PNG and GeoTiff case. Is `bytes` an option or is that a keyword?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
```diff
@@ -61,4 +96,15 @@ trait TupleProtoBuf {
...
+
+  implicit def tupleSpaceTimeKeyImageProtoBufCodec =
+    new ProtoBufCodec[(SpaceTimeKey, Array[Byte]), ProtoTuple] {
+    def encode(tuple: (SpaceTimeKey, Array[Byte])): ProtoTuple =
+      ProtoTuple(
+        spaceTimeKey = Some(spaceTimeKeyProtoBufCodec.encode(tuple._1)),
+        imageBytes = ByteString.copyFrom(tuple._2))
+
+    def decode(message: ProtoTuple): (SpaceTimeKey, Array[Byte]) =
+      throw new Exception("This ProtoBuf message cannot be decoded")
```
echeipesh: This could be decoded easily actually. Somebody will hit this, best to say its not implemented rather that it isn't done.

geopyspark/geotrellis/protobufcodecs.py
```diff
@@ -273,6 +273,42 @@ def create_partial_tuple_decoder(key_type):
...
+    if key_decoder == "ProjectedExtent":
+        return (from_pb_projected_extent(tup.projectedExtent), image_bytes)
+    elif key_decoder == "TemporalProjectedExtent":
+        return (from_pb_temporal_projected_extent(tup.temporalProjectedExtent), image_bytes)
+    elif key_decoder == "SpatialKey":
+        return (from_pb_spatial_key(tup.spatialKey), image_bytes)
+    else:
+        return (from_pb_space_time_key(tup.spaceTimeKey), image_bytes)
+
+def create_partial_image_rdd_decoder(key_type):
```
echeipesh: The image comment again, `create_partial_bytes_rdd_decoder` seems more descriptive, open to other suggestions for this name though. (+ all related methods)

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
jbouffard: I chose `imageBytes` because it's the `bytes` of both the PNG's and GeoTiffs which are both images. I could think of something else, though.

Which `bytes` are you referring to? `bytes` in ProtoBuf is optional.

geopyspark/geotrellis/protobufcodecs.py
jbouffard: The reason I chose that name is because right now, it just deals with PNG and soon GeoTiff bytes. However, I can change this if it's not clear or confusing.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
jbouffard: That would make things a bit easier. Is this something that should be implemented in this PR? Or just something to consider for later?

jbouffard:
@echeipesh How about using `file` instead of `image` in the naming schemes? The names: `fileBytes' and `create_partial_file_rdd_decoder` can be used in place of `imageBytes` and `create_partial_image_rdd_decoder`, respectively.

echeipesh:
`..._file_rdd` is also weird. I'm going to withdraw that comment and just not worry about it in this case. These are fairly internal functions after all.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
echeipesh: Let me actually throw it into the compiler to double check. I see that I missed a type parameter in my mockup. If it works it would remove a lot of the code introduced in this PR, so it should really be here.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TupleProtoBuf.scala
echeipesh: Right the problem is the `projectedExtent = Some(keyCodec.encode(tuple._1))` where we expect to switch which field we set based on type. There is a way to work around this with lenses but yeah, that doesn't make sense for this PR.

Issue #150, opened 2017-05-01
TiledRasterRDD.pyramid does not work on layers that are not power of 2


echeipesh:
The `pyramid` function delegates to [`geotrellis.spark.pyramid.Pyramid.levelStream`](https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/pyramid/Pyramid.scala#L171) which returns the input RDD as base level.

Therefore if the target `TiledRasterRDD` is not already in power of 2 layout that matches the base zoom level it will be incorrectly returned as the base level and higher level of the pyramid will have malformed output.

Originally there was logic to handle this case that [got removed](https://github.com/locationtech-labs/geopyspark/commit/77487f728b3354cb1944b98509ff601c85e0f13e). Unless there is a better API choice to be made here it should probably be reinstated.

jbouffard:
The reason it was removed was because after I talked with @lossyrob about this, it was decided that it was up to the user to make sure that their inputs are in the correct format. I could add warnings, though, that show up when a bad input is trying to be used.

PR #316, opened 2017-07-03
Non-Power-of-Two Pyramiding

jamesmcclain:
The `TiledRasterLayer` type now has a `pyramid_non_power_of_two` method.  The error message that results from using `pyramid` with non-power-of-two input now gives a pointer to the new method.

Fixes locationtech-labs/geopyspark#150

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
@@ -333,21 +333,16 @@ class SpatialTiledRasterRDD(
     val rdd2 = rdd.mapValues({ tile => tile.resample(cols, rows, method) })
 
     val metadata = rdd.metadata
-    val metadata2 = TileLayerMetadata(
-      metadata.cellType,
-      LayoutDefinition(
-        metadata.extent,
-        TileLayout(
-          metadata.layout.tileLayout.layoutCols,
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
```diff
@@ -316,6 +322,36 @@ class SpatialTiledRasterRDD(
...
+    row_power: Int,
+    resampleMethod: String
+  ): TiledRasterRDD[SpatialKey] = {
+    val method: ResampleMethod = TileRDD.getResampleMethod(resampleMethod)
+    val cols = 1<<col_power
+    val rows = 1<<row_power
+    val rdd2 = rdd.mapValues({ tile => tile.resample(cols, rows, method) })
+
+    val metadata = rdd.metadata
+    val metadata2 = TileLayerMetadata(
```
jbouffard: What do you think about using `copy` to create a new instance of `TileLayerMetadata`?

```scala

val metadata = rdd.metadata
val metadata2 = metadata.copy(layoutDefinition = LayoutDefinition(metadata.extent, TileLayout(metadata.layout.tileLayout.layoutCols, metadata.layout.tileLayout.layoutRows, cols, rows))

jamesmcclain: I am enthused.

PR #327, opened 2017-07-07
No Data Value Fix

### LINKED ISSUE #320: NoData values are not ending up in numpy array correctly
screen shot 2017-07-05 at 1 42 43 pm

In this example, I have the chatta_rdd, which is defined as

```python
MAX_ZOOM = 12
query_rdd = catalog.query(
    pysc, LayerType.SPATIAL, 
    "s3://azavea-datahub/catalog", nlcd_layer_name, 
    MAX_ZOOM, query_geom=chatta_poly)

chatta_rdd = query_rdd.convert_data_type("int8").cache()
```

When I mask it, and get a numpy array out of it, the nodata values (masked values) are set to 0. However based on the nodata they should be set to -128, correct?

jbouffard:
This PR fixes an error that occurs when encoding/decoding no data values. The cause of this issue has to do with how ProtoBuf encoded the no data values and how numpy represented these new values.

For more information, see the issue this PR resolves #320

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TileProtoBuf.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TileProtoBuf.scala
@@ -130,27 +130,10 @@ trait TileProtoBuf {
       val ct = messageToCellType(messageCellType)
 
       message.cellType.get.dataType.toString match {
-        case "BIT" =>
-            ArrayTile(message.uint32Cells.toArray, message.cols, message.rows).interpretAs(ct)
-
-        case ("BYTE" | "SHORT") =>
-          if (messageCellType.hasNoData)
-            ArrayTile(message.sint32Cells.toArray, message.cols, message.rows)
-              .interpretAs(ct)
-              .map { x => if (x == Int.MinValue) messageCellType.nd.toInt else x }
-          else
-            ArrayTile(message.sint32Cells.toArray, message.cols, message.rows).interpretAs(ct)
-
-        case ("UBYTE" | "USHORT") =>
-          if (messageCellType.hasNoData)
-            ArrayTile(message.uint32Cells.toArray, message.cols, message.rows)
-              .interpretAs(ct)
-              .map { x => if (x == Int.MaxValue) messageCellType.nd.toInt else x }
-          else
-            ArrayTile(message.uint32Cells.toArray, message.cols, message.rows).interpretAs(ct)
-
-        case "INT" =>
-            ArrayTile(message.sint32Cells.toArray, message.cols, message.rows).interpretAs(ct)
+        case ("BYTE" | "SHORT" | "INT") =>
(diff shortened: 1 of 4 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/protobufs/TileProtoBuf.scala
```diff
@@ -126,13 +126,31 @@ trait TileProtoBuf {
...
-        case ("UBYTE" | "USHORT" | "BIT") =>
-          ArrayTile(message.uint32Cells.toArray, message.cols, message.rows).interpretAs(ct)
+        case "BIT" =>
+            ArrayTile(message.uint32Cells.toArray, message.cols, message.rows).interpretAs(ct)
+
+        case ("BYTE" | "SHORT") =>
+          if (messageCellType.hasNoData)
+            ArrayTile(message.sint32Cells.toArray, message.cols, message.rows)
+              .interpretAs(ct)
+              .map { x => if (x == Int.MinValue) messageCellType.nd.toInt else x }
```
echeipesh: Was this map step needed? InterpretAs should have taken care of this.

echeipesh: Ah, ArrayTile comes in with constant no data value. If you use RawArrayTile it will come in with no nodata value and you can just use interpretAs, should not even need the if statement

jbouffard:
@echeipesh Cleaned up the code and added a test.

geopyspark/geotrellis/protobufcodecs.py
```diff
@@ -27,28 +28,28 @@
...
 
 
-def from_pb_tile(tile, data_type=None):
+def from_pb_tile(tile, no_data_value=None, data_type=None):
     if not data_type:
         data_type = _mapped_data_types[tile.cellType.dataType]
 
     if data_type == 'BIT':
-        cells = np.int8(tile.uint32Cells[:]).reshape(tile.rows, tile.cols)
+        cells = np.int8([no_data_value if x == SCALA_MAX_INT else x for x in tile.uint32Cells])
```
echeipesh: It should be possible to avoid these if during encoding the tiles are converted to "raw" form of their cell type, which will expose the NoData values as the actual values rather than `Int.MinValue` and `Double.NaN`

here is some proof:
```scala
cala> val tile = ArrayTile.empty(ShortConstantNoDataCellType, 4,4)
tile: geotrellis.raster.MutableArrayTile = ShortConstantNoDataArrayTile([S@2091c7da,4,4)

scala> tile.toArray()
res6: Array[Int] = Array(-2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648, -2147483648)

scala> tile.interpretAs(ShortCellType).toArray()
res7: Array[Int] = Array(-32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768, -32768)
```

PR #328, opened 2017-07-07
to_geotiff_rdd

### LINKED ISSUE #278: Add .to_geotiff_rdd() to RasterRDD and TiledRasterRDD
This should use the Scala code to create RDD of `(SpatialKey, bytes)` or `(SpaceTimeKey, bytes)` of the GeoTiff bytes.

jbouffard:
This PR adds the `to_geotiff_rdd` method to RasterLayer and TiledRasterLayer. When called, this will return a RDD[(K, bytes)] where the bytes the GeoTiff bytes of the raster and the K is the corresponding key of the GeoTiff.

This PR is based on another, open PR #327 

This PR resolves #278

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
@@ -200,15 +200,12 @@ abstract class RasterRDD[K: ClassTag] extends TileRDD[K] {
     withRDD(rdd.mapValues { multibandTile => multibandTile.subsetBands(bands.asScala) })
 
   def toGeoTiffRDD(
-    storageMethod: String,
-    rowsPerStrip: Int,
-    tileDimensions: java.util.ArrayList[Int],
+    storageMethod: StorageMethod,
     compression: String,
     colorSpace: Int,
     headTags: java.util.Map[String, String],
     bandTags: java.util.ArrayList[java.util.Map[String, String]]
   ): JavaRDD[Array[Byte]] = {
-    val storage = TileRDD.getStorageMethod(storageMethod, rowsPerStrip, tileDimensions)
     val tags =
       if (headTags.isEmpty || bandTags.isEmpty)
         Tags.empty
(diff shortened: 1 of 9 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/RasterRDD.scala
```diff
@@ -180,6 +199,58 @@ abstract class RasterRDD[K: ClassTag] extends TileRDD[K] {
...
+    val options = GeoTiffOptions(
+      storageMethod,
+      TileRDD.getCompression(compression),
+      colorSpace,
+      None)
+
+    toGeoTiffRDD(tags, options)
+  }
+
+  def toGeoTiffRDD(
```
echeipesh: This guy can move up to `TileRDD` along with the related abstract methods since no type tie them to either `RasterRDD` or `TiledRasterRDD`

geopyspark/geotrellis/layer.py
```diff
@@ -232,6 +266,55 @@ def to_png_rdd(self, color_map):
 
         return create_python_rdd(self.pysc, result, ser)
 
+    def to_geotiff_rdd(self,
+                       storage_method=StorageMethod.STRIPED,
+                       rows_per_strip=None,
+                       tile_dimensions=(256, 256),
```
echeipesh: The default option scala side is [striped](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffOptions.scala#L39). I don't know that this is a wrong choice to go with tiled here, curious what your thoughts where.

echeipesh: Sorry, misread the structure here, the striped is the default. I am no longer curious : )

echeipesh:
Aside from code duplication which can be fixed by moving a method to the superclass this looks really good. Excited to try to save out chunk of NLCD from the azavea datahub.

jbouffard:
@echeipesh Thanks! Yeah, I think this is going to be a pretty useful feature. I'm definitely going to use it to create data sets for tests and examples.

PR #329, opened 2017-07-10
Band Selection

### LINKED ISSUE #324: Work With Specific Bands of a MultibandTile
Right now, operations, analysis, and display are performed on all bands. However, as @rossbernet pointed out, there are times when the focus is just on a few bands, and not all of them. Therefore, it would be beneficial if we had some API to address this before releasing 0.2.

Here are a few possible implementations of this API:

- Have the option available when reading

```python
  # Only the first and third band will be read in from the files

  raster_layer = get(pysc, "Spatial", "/path/to/my/files", bands_of_interest=[0, 2])
  tiled_layer = query(pysc, "Spatial", "file://path/to/my/catalog", "test-layer", 10, bands_of_interest=[0, 2])
```

- A special filter method on the layers

```python
  # Only the first band will taken

  filtered_raster_layer = raster_layer.select_bands(0)
  filtered_tiled_layer = tiled_layer.select_bands(0)
```

- Selecting bands during operations

```python
  # Polygonal min on only the second and third bands

  polygonal_min = tiled_layer.polygonal_min(my_geom, int, bands_of_interest=[1, 2])
```

Of the above three, I like the first two the most. The latter would require too much rewriting of existing methods/function, in my opinion. @echeipesh @jpolchlo What do you guys think?

jbouffard:
This PR adds to new methods to `RasterLayer` and `TiledRasterLayer` that allows a user to select certain bands or to map over each `Tile` within the layer. These are `bands` and `map_tiles`, respectively. This will allow the user to work with specific bands without having to serialize the wrapped `RDD` to Python itself.

This PR resolves #324

### DIFF
--- a/geopyspark/geotrellis/layer.py
+++ b/geopyspark/geotrellis/layer.py
@@ -255,13 +255,32 @@ class RasterLayer(CachableLayer):
             ``extent`` and ``layout`` must both be defined if they are to be used.
 
         Returns:
-            :class:`~geopyspark.geotrellis.rdd.TiledRasterLayer`
+            :class:`~geopyspark.geotrellis.rdd.RasterLayer`
         """
 
         return self.tile_to_layout(self.collect_metadata(extent, layout, crs, tile_size),
                                    resample_method)
 
     def bands(self, band):
+        """Select a subsection of bands from the ``Tile``\s within the layer.
+
+        Note:
+            There could be potential high performance cost if operations are performed
+            between two sub-bands of a large data set.
+
+        Note:
+            Due to the natue of GeoPySpark's backend, if selecting a band that is out of bounds
+            then the error returned will be a ``py4j.protocol.Py4JJavaError`` and not a normal
+            Python error.
+
+        Args:
+            band (int or tuple or list or range): The band(s) to be selected from the ``Tile``\s.
+                Can either be a single ``int``, or a collection of ``int``\s.
+
+        Returns:
+            :class:`~geopyspark.geotrellis.rdd.RasterLayer` with the selected bands.
+        """
+
         if isinstance(band, range):
             band = list(band)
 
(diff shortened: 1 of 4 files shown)


geopyspark/geotrellis/layer.py
```diff
@@ -255,12 +255,63 @@ def to_tiled_layer(self, extent=None, layout=None, crs=None, tile_size=256,
...
 
         Returns:
-            :class:`~geopyspark.geotrellis.rdd.TiledRasterLayer`
+            :class:`~geopyspark.geotrellis.rdd.RasterLayer`
         """
 
         return self.tile_to_layout(self.collect_metadata(extent, layout, crs, tile_size),
                                    resample_method)
 
+    def bands(self, band):
```
echeipesh: It seemed initially that this should have been called `subsetBands` to mirror GT method name. But for the purpose of the python API `bands` seems plenty clear.

geopyspark/geotrellis/layer.py
```diff
@@ -255,12 +255,63 @@ def to_tiled_layer(self, extent=None, layout=None, crs=None, tile_size=256,
...
+        """Select a subsection of bands from the ``Tile``\s within the layer.
+
+        Note:
+            There could be potential high performance cost if operations are performed
+            between two sub-bands of a large data set.
+
+        Note:
+            Due to the natue of GeoPySpark's backend, if selecting a band that is out of bounds
+            then the error returned will be a ``py4j.protocol.Py4JJavaError`` and not a normal
+            Python error.
```
echeipesh: Is it at all possible to catch and re-throw a more useful error message? This is bound to happen pretty often.

geopyspark/tests/band_selection_test.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_band_range(self):
+        actual = self.raster_rdd.bands(range(0, 3)).to_numpy_rdd().first()[1]
+        expected = np.array([self.band_1, self.band_2, self.band_3])
+
+        self.assertTrue((expected == actual.cells).all())
+
+    def test_map_tiles_func(self):
+        def test_func(tile):
+            cells = tile.cells
+            return Tile((cells[0] + cells[1]) / cells[2], tile.cell_type, tile.no_data_value)
```
echeipesh: Making this function `Tile => Tile` is clean from perspective of API but awkward from perspective of the user because the actual work needs to be re-wrapped in `Tile` with no data value and cell type. Those arguments are not going to change in 90% of the use-cases (I expected).

Perhaps this can be solved by adding the `map_cells` function which  expects as input `(cells, nd) => cells`?

geopyspark/geotrellis/layer.py
jbouffard: I was never to get it to work. There is currently an open issue for it https://github.com/bartdag/py4j/issues/153

geopyspark/tests/band_selection_test.py
jbouffard: Yeah, I think that'd make things a bit easier for the user.

geopyspark/geotrellis/layer.py
echeipesh: Drat, thats too much potential confusion to leave hanging around. We could return `Either[RasterLayer[K], Exception]` or `Either[RasterLayer[K], String]`. 

Both `Left` and `Right` have `.value` so python caller could call `result.value` and match on the type. Since the call will go through reflection, we know we will get something back.
Maybe this would even warrant making a py4j decoder to go from scala `Exception` to python error.

echeipesh:
Going to merge this since both of the things we talked about would not impact the API added in this PR.

PR #334, opened 2017-07-10
Batching TMS RDD server

jpolchlo:
Display of RDD data through the TMS server was a bit slow.  Every request was triggering a separate call to `lookup()` on the pyramid's RDDs, which could be quite slow for larger data sets.  To overcome this, a batching method was implemented in this PR to replace a `lookup` with a `filter`.  Aggregating lookup requests significantly reduces the number of iterations over the RDDs and yields much more responsive display.

The new TMS server uses an Akka actor system to allow requests to queue up.  Using the Akka scheuler, the queue is polled at a set interval (currently 150ms).  Polling avoids "warm-up" time, where the first tile requested is served separately before queue is allowed to fill.

NOTE: This PR relies on geotrellis/geonotebook#11 to realize the gains in parallelism.

- [X] Proof of concept implementation
- [x] Startup Akka system upon binding
- [x] Ensure proper shutdown of TMS server
- [x] Ensure correct function with other TMS backends


echeipesh:
Can confirm the RDD layer tiles are loading in parallel:
screen shot 2017-07-11 at 10 37 06 am

echeipesh:
Have been using this PR and it behaves. It needs to remove commented code and maybe break out the `TmsRoutes.scala`, that file is getting a little packed for scala convention.

PR #343, opened 2017-07-10
Docstrings Update

jbouffard:
This PR updates and expands the various docstrings in the `geopyspark.geotrellis` sub-package. Since almost all of the modules the user needs to use are located here, docstrings outside this sub-package were thus not the focus of this PR or of 0.2.

This PR is based on two, already open PRs: #342 and #341

### DIFF
--- a/geopyspark/__init__.py
+++ b/geopyspark/__init__.py
@@ -12,10 +12,12 @@ from pyspark.serializers import AutoBatchedSerializer
 
 
 def map_key_input(key_type, is_boundable):
-    """Gets the mapped GeoTrellis type from the `key_type`.
+    """Gets the mapped GeoTrellis type from the ``key_type``.
+
     Args:
         key_type (str): The type of the ``K`` in the tuple, ``(K, V)`` in the RDD.
         is_boundable (bool): Is ``K`` boundable.
+
     Returns:
         The corresponding GeoTrellis type.
     """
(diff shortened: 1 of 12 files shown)


echeipesh:
Looking over what has changed, not what hasn't, this looks good.
I have the following questions:

1. I see that the notation we're using is evolving as we're updating the docs. There used to be `:cls:` now things are `:class:` the way things are quoted are changing. Is there a standard reference people can use when creating doc strings to minimize this kind of flutter as we update?

2. Related to above: Is `:class:` appropriate for `Enums`?

3. What is the convention for `Returns`? Sometimes it s naked class name like `RDD` and sometimes its linked like `:class:`~geopyspark.geotrellis.tms.TMSServer`

jbouffard:
@echeipesh 

1. There is. The style all docstrings are written in is, Google's Python Style DocString, also know as [napolean](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). `:cls:` and `:class:` are actually not apart of napolean. Rather, these notations are used to create links to other parts of the docs once published. This building and generation of the docs is handled by an application called, Sphinx. Sphinx itself has a good guide on how use these features called, [domains](http://www.sphinx-doc.org/en/stable/domains.html)

2. `:class:` is appropriate for `Enum` because they are represented as `class`es in the code and Sphinx makes no distinction between a `class` and an `Enum` in its domains.

3. For returns, I tend to use `:class:`, `:obj:`, etc. when the return value is an instance of something within GeoPySpark. If what's being returned is either an instance of another library or apart of the standard Python, then I just use `RDD`, `[int]`, etc. And if nothing is returned or the docstring is for a class, then I just leave out the `Return` statement.

PR #352, opened 2017-07-12
User-defined Python rendering functions for TMS display

### LINKED ISSUE #281: Accept TileRender in tms functions as an argument

jpolchlo:
There may be cases where a user wishes to define a custom rendering function in Python that will transform a numpy array into a bytes representation of an image, the input array having shape (#bands, #rows, #cols).  This PR enables that capability.  Given a function from nparray to image bytes, we wrap that function in a `TileRender` object and pass it to either `TMSServer.rdd_tms_server` or `.s3_catalog_tms_server`.  The results will be slower to display, but adds a measure of flexibility.

_As a note, it appears that Py4J exception reporting when Scala invokes a Python function that results in errors is the pits.  One will either encounter a NullPointerException or receive the cryptic error message `py4j.Py4JException: An exception was raised by the Python Proxy. Return Message: x`.  In these cases, more explicit testing of the rendering function within Python itself is warranted._

Fixes #281

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/tms/RenderFunction.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/tms/RenderFunction.scala
@@ -4,9 +4,12 @@ import geotrellis.raster._
 import geotrellis.raster.render._
 
 trait TileRender {
-  def render(tile: Tile): Array[Byte]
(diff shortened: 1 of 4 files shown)


jpolchlo:
A demo notebook can be found at https://gist.github.com/jpolchlo/ac54aa8f1cd0cb807e4dc932686b0323

echeipesh:
Superseded by #363

PR #363, opened 2017-07-17
API updates and compositing callbacks for TMS

### LINKED ISSUE #279: Compositing JVM TMS endpoint
We want to be able to read multiple layers and render them on python side as list of numpy arrays, given in the same order as the list of sources.

Bonus: Mixing catalog tiles with RDD tiles.

jpolchlo:
On the way to GeoPySpark 0.2, the TMS API needed cleaning up.  On the back of locationtech/geotrellis#2223, we are able to move toward using a simpler URI-based scheme for identifying catalog layers, and in the process, it is sensible to simplify the construction of TMS servers.

See https://gist.github.com/jpolchlo/70798bac6a71235866841bda99c262cb for an example of the new building method.

At this time, we also have a good opportunity to provide a second callback rendering pathway that allows for multiple tile sources to be combined through a single compositing function.  This is generally expected to be used from the Python side; as such, there is no "sensible default" Scala implementation provided.

Fixes #279


geopyspark/geotrellis/tms.py
```diff
@@ -38,83 +46,138 @@ def render(self, cells, cols, rows): # return `bytes`
...
+        pysc._gateway.start_callback_server()
+
+    def set_handshake(self, handshake):
+        self.server.set_handshake(handshake)
+        self.handshake = handshake
 
     @classmethod
-    def rdd_tms_server(cls, pysc, pyramid, color_map):
-        """Creates a ``TMSServer`` from a ``Pyramid`` instance.
+    def build(cls, pysc, *args, **kwargs):
```
jbouffard: What exactly is the purpose of having the parameters be `*args` and `**kwargs`? I think it makes things more difficult to use.

```python
def build(cls, pysc, source, display_method)
```

How does the above differ from what is there already?

jpolchlo: The point of this is that the user can provide a list of different tile sources to the `build` method.  So, we need `*args`.  The user must also provide a display method, but because there is a distinction between a renderer (takes a single tile input) and a compositer (takes a list of tiles as input), and there's no way to determine which type of function was supplied, we need a named argument.  These cannot come before the `*args`, and the argument list will not play nice with default parameters, so we're stuck with `**kwargs`, too.

jbouffard: Would something like this work for the `*args`?

```python
def build(cls, pysc, source, *args, renderer=None, compositor=None)
```

That way the user knows that there needs to be at least one `source` parameter with the option of passing in more to the method. I'm also still not sure why the `renderer` and the `compositor` can't just be named parameters with a default of `None` for both. With the distinction between the two being made in the documentation. The user already has to specify which one they're using, this just makes it easier to discover the argument names.

jpolchlo: You're right about the `render` and `composite` args.  My mistake.  Thought I checked if that would work or not.

Maybe the right thing to do here is to allow the `source` argument to be a regular value or a list?  That would also disambiguate the nature of the display parameter, then we can move to a single argument for the display method as well.

jbouffard: I think making it either a single value or a list is a good idea. It'll make things much more clear and would provide an easy way to check the inputs to make sure they're correct.

Issue #348, opened 2017-07-11
Numpy Array Type to Tile CellType


jbouffard:
There is no function as of right now that goes from `np.ndarray.dtype` => `Tile.cell_type`. This would help when creating `Tile`s from `np.ndarray`s.

jbouffard:
Resolved in #373

PR #373, opened 2017-07-19
Tile Class Methods

### LINKED ISSUE #356: Tile.from_numpy_array Class Method
It would be nice to have a class method for `Tile` that would construct an instance of it from a `numpy.ndarray` and an optional `no_data_value`. This would make it easier for users as they don't have to worry about setting `cell_type`.

It could look something like this:

```python

arr = np.zeros((2, 16, 16), dtype=int)

Tile.from_numpy_array(arr, no_data_value=-1)
Tile(cells=arr, cell_type="INT", no_data_value=-1)
```

jbouffard:
This PR makes `Tile` a class and adds a `staticmethod` and `classmethod` to it, `dtype_to_cell_type` and `from_numpy_array`, respectively. By adding these methods, it will now be easier for users to create `Tile` instances without having to worry about choosing the correct `cell_type` for the `Tile`.

This PR resolves #356 #348

### DIFF
--- a/geopyspark/geotrellis/__init__.py
+++ b/geopyspark/geotrellis/__init__.py
@@ -23,8 +23,8 @@ def deprecated(func):
     return new_func
 
 
-Tile = namedtuple("Tile", 'cells cell_type no_data_value')
-"""Represents a raster in GeoPySpark.
+class Tile(namedtuple("Tile", 'cells cell_type no_data_value')):
+    """Represents a raster in GeoPySpark.
 
     Note:
         All rasters in GeoPySpark are represented as having multiple bands, even if the original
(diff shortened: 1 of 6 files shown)


geopyspark/geotrellis/__init__.py
```diff
@@ -33,12 +33,76 @@ def new_func(*args, **kwargs):
...
+        elif name == 'int16':
+            return 'SHORT'
+        elif name == 'uint16':
+            return 'USHORT'
+        elif name == 'int32':
+            return 'INT'
+        elif name in ['uint32', 'float16', 'float32']:
+            return 'FLOAT'
+        elif name in ['int64', 'uint64', 'float64']:
+            return 'DOUBLE'
```
echeipesh: It's a shame we couldn't use just the numpy convention for the celltypes

PR #376, opened 2017-07-19
Renamed Scala Classes And Objects to Match Python Naming Scheme

### LINKED ISSUE #371: Rename Scala Classes and Objects
Currently, the names of the Scala classes and objects used on the Python side still retain the old naming convention of having `RDD` in their name. These names should be changed so that `RDD` becomes `Layer` in order to match the Python API. This'll make the code less confusing to read for someone not familiar with the development history of GeoPySpark.

jbouffard:
This PR renames the Scala classes and objects to match the Python naming scheme which will make it easier for people to read GeoPySpark's code.

This PR resolves #371


jbouffard:
Superseded by #379

Issue #370, opened 2017-07-19
Add .is_global_layer to TiledRasterLayer


echeipesh:
This is something that can be inferred from relationship between `zoom_level` and `metadata`, some code in https://github.com/locationtech-labs/geopyspark/issues/369 will have to make decisions based on this so we may as well exposed it as API for clarity.

jbouffard:
Our final version of the API does not need this feature and thus this issue shall be closed.

PR #379, opened 2017-07-20
Add GlobalLayout and LocalLayout

echeipesh:
Abstracts over layout scheme before CRS is known


geopyspark/geotrellis/converters.py
```diff
@@ -21,4 +22,19 @@ def convert(self, object, gateway_client):
...
+
+    def convert(self, obj, gateway_client):
+        if isinstance(obj, GlobalLayout):
+            cls = JavaClass("geopyspark.geotrellis.GlobalLayout$", gateway_client)
+            return cls().apply(obj.tile_size, obj.zoom, obj.threshold)
+        elif isinstance(obj, LocalLayout):
+            cls = JavaClass("geopyspark.geotrellis.LocalLayout$", gateway_client)
+            return cls().apply(obj.tile_size)
+        else:
+            raise TypeError("Could not convert {} to geotrellis.raster.LayouType".format(obj.sampleType))
```
jpolchlo: LayouType => LayoutType

geopyspark/geotrellis/__init__.py
```diff
@@ -234,6 +234,45 @@ def _asdict(self):
...
+                                that is tolerated to snap to the lower-resolution zoom level.
+                                For example, if this paramter is 0.1, that means we're willing to downsample
+                                rasters with a higher resolution in order to fit them to some zoom level Z,
+                                if the difference is resolution is less than or equal to 10% the difference
+                                between the resolutions of zoom level Z and zoom level Z+1.
+
+Returns:
+    :obj:`~geopyspark.geotrellis.GlobalLayout`
+"""
+GlobalLayout.__new__.__defaults__ = (256, None, 0.1)
```
jpolchlo: Honest question: always square?  What if we're dealing with ingested data with non-square tiles?

echeipesh: `ZoomedLayoutScheme` is coded to always be square though. dealing with ingested non square tiles is a pretty good reason to expose tileCols, tileRows in `LocalLayout` for that case eventually when you read in already tiled dataset and you just need to give it a key.

geopyspark/geotrellis/layer.py
```diff
@@ -480,46 +504,50 @@ def collect_metadata(self, extent=None, layout=None, crs=None, tile_size=256):
...
+                Either EPSG code, well-known name, or a PROJ.4 string.
+            layout (
+                :obj:`~geopyspark.geotrellis.GlobalLayout` or
+                :obj:`~geopyspark.geotrellis.LocalLayout`, optional
+            ): Target raster layout when reprojecting.
             resample_method (str or :class:`~geopyspark.geotrellis.constants.ResampleMethod`, optional):
                 The resample method to use for the reprojection. If none is specified, then
                 ``ResampleMethods.NEAREST_NEIGHBOR`` is used.
 
         Returns:
```
jbouffard: I know that we wanted to have just one method handling all cases, but do you think it might be better to split the `reproject` method into two, separate operations? One would never sample past the tiles' boundaries and the other would. That might be better than having a method return one of two things depending on its inputs.

geopyspark/geotrellis/layer.py
```diff
@@ -992,81 +1020,50 @@ def tile_to_layout(self, layout_definition, resample_method=ResampleMethod.NEARE
...
-        result = self.srdd.pyramid(start_zoom, end_zoom, ResampleMethod(resample_method).value)
+            raise TypeError("Could not retile from the given layout", layout)
 
-        return Pyramid([TiledRasterLayer(self.pysc, self.layer_type, srdd) for srdd in result])
+        return TiledRasterLayer(self.pysc, self.layer_type, srdd)
 
-    def pyramid_non_power_of_two(self, col_power, row_power, end_zoom, start_zoom=None, resample_method=ResampleMethod.NEAREST_NEIGHBOR):
-        """Creates a ``Pyramid`` instance. This will work even if the tiles' sizes are not
-        multiples of 2.
+    def pyramid(self, resample_method=ResampleMethod.NEAREST_NEIGHBOR):
```
jbouffard: Should the user be allowed to specify a `max_zoom` and `min_zoom`?

geopyspark/geotrellis/layer.py
echeipesh: I thought about that but couldn't really think of a good name for two methods that wouldn't be confusing. The "buffered" reproject is GeoTrellis nomenclature and doesn't hold a lot of meaning outside of that. I decided that buffering to a layout held the meaning of buffered reproject while buffering without a layout could only mean you're doing reprjection on per tile basis.

@rossbernet for second opinion on API

geopyspark/geotrellis/layer.py
echeipesh: In the end trying to implement that didn't make sense. Pyramiding to zoom 0 is cheap and its hard to imagine why somebody might want to stop short. If that is really wanted, they can always discard the levels they're not interested it.

Pyramiding from zoom level other than the one you're at should require a re-tile operation, might as well make that explicit so the user gets to decide how it needs to be done.

echeipesh:
Its worth pointing out explicitly that the implemented `pyramid` function will pyramid local and global layers in exactly the way you would expect. Examining if a layer has a zoom level or not to determine whether to apply global or local pyramid logic.

However once the local layout layer has been put in a pyramid it gains a zoom level (and all of the other levels in the pyramid) and its no longer super easy to tell what kind of layer it is.

You can give the local pyramid to a TMS server and it will display, of course not lining up with base map. I'm not sure how concerned to be with this because it ends up being a super obvious mistake, but its a short-coming in current API.

geopyspark/geotrellis/layer.py
jbouffard: What about `reproject` and `reproject_to_layout`? Where the first doesn't do the buffered reproject and doesn't require a `layout` while the latter does.

geopyspark/geotrellis/layer.py
jbouffard: I see. Those are good points. Do you think a method should be added to `Pyramid` that lets the user select certain zoom levels?

Something like this:

```python

def select_levels(self, min_zoom=0, max_zoom=None):
    max_zoom = max_zoom or self.max_zoom
```

jbouffard:
@echeipesh Would implementing #370 help alleviate the possible confusion in the `pyramid` API? There could be a simple check down when using the TMS server that'll make sure the layer is in a `GlobalLayout` and fail if that's not the case.

echeipesh:
Perhaps that, its not clear how that decision would be made though, because its conceivable that you have a local layer after pyramid that has a power of two layout cols/rows and a zoom level. Perhaps the thing to do is to carry the `layout_type` parameter around as a field after initial tiling is applied, so it could always be inspected.

PR #385, opened 2017-07-20
Updated TiledRasterLayer's Reproject Method

jbouffard:
This PR updates the `reproject` method for `TiledRasterLayer` by having it accept various values for the `layout` including `LocalLayout` and `GlobalLayout`.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
@@ -100,31 +100,25 @@ abstract class TiledRasterRDD[K: SpatialComponent: JsonFormat: ClassTag] extends
     crs: String,
     resampleMethod: ResampleMethod
   ): TiledRasterRDD[K] = {
-    val layoutDefinition = Right(LayoutDefinition(extent.toExtent, layout.toTileLayout))
+    val layoutDefinition = LayoutDefinition(extent.toExtent, layout.toTileLayout)
 
     reproject(layoutDefinition, TileRDD.getCRS(crs).get, getReprojectOptions(resampleMethod))
   }
 
   def reproject(
-    scheme: String,
-    tileSize: Int,
-    resolutionThreshold: Double,
+    layoutType: LayoutType,
     crs: String,
     resampleMethod: ResampleMethod
   ): TiledRasterRDD[K] = {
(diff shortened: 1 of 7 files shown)


geopyspark/geotrellis/layer.py
```diff
@@ -892,54 +892,61 @@ def convert_data_type(self, new_type, no_data_value=None):
             return TiledRasterLayer(self.pysc, self.layer_type,
                                     self.srdd.convertDataType(CellType(new_type).value))
 
-    def reproject(self, target_crs, extent=None, layout=None, scheme=LayoutScheme.FLOAT, tile_size=256,
-                  resolution_threshold=0.1, resample_method=ResampleMethod.NEAREST_NEIGHBOR):
+    def reproject(self, target_crs, layout=LocalLayout(),
```
echeipesh: Default here should be `None`, trying to match the layout there is.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterRDD.scala
```diff
@@ -100,31 +100,25 @@ abstract class TiledRasterRDD[K: SpatialComponent: JsonFormat: ClassTag] extends
...
-    tileSize: Int,
-    resolutionThreshold: Double,
+    layoutType: LayoutType,
     crs: String,
     resampleMethod: ResampleMethod
   ): TiledRasterRDD[K] = {
-    val _crs = TileRDD.getCRS(crs).get
+    val convertedCRS = TileRDD.getCRS(crs).get
+    val layoutDefinition =
+      layoutType.layoutDefinition(convertedCRS, rdd.metadata.extent, rdd.metadata.cellSize)
```
jbouffard: Would doing `rdd.metadata.extent.reproject(convertedCRS)` work?

echeipesh: No, that won't do the trick by itself. The reprojection could bow the extent where just reproejcting the envelope won't actually cover all the reprojected cells. I think we may have densified the envelope to reproject it in the past, but I'll have to double check that.

jbouffard: I see. In the meantime then, I fix the python code.

jbouffard:
Superseded by #390

PR #387, opened 2017-07-20
RasterRDD.tile_to_layout with LayoutType

echeipesh:


geopyspark/geotrellis/layer.py
```diff
@@ -541,25 +541,34 @@ def cut_tiles(self, layer_metadata, resample_method=ResampleMethod.NEAREST_NEIGH
...
 
-    def tile_to_layout(self, layer_metadata, resample_method=ResampleMethod.NEAREST_NEIGHBOR):
+    def tile_to_layout(self, layout=LocalLayout(), resample_method=ResampleMethod.NEAREST_NEIGHBOR):
         """Cut tiles to layout and merge overlapping tiles. This will produce unique keys.
 
         Args:
-            layer_metadata (:class:`~geopyspark.geotrellis.Metadata`): The
-                ``Metadata`` of the ``RasterLayer`` instance.
+            layout (
+                :class:`~geopyspark.geotrellis.Metadata` or
```
jbouffard: `TiledRasterLayer` should be added to the list of accepted types.

geopyspark/geotrellis/layer.py
```diff
@@ -541,25 +541,34 @@ def cut_tiles(self, layer_metadata, resample_method=ResampleMethod.NEAREST_NEIGH
...
+                :obj:`~geopyspark.geotrellis.GlobalLayout` or
+                :obj:`~geopyspark.geotrellis.LocalLayout`
```
jbouffard: We should also signify that the `layout` parameter is optional.

geopyspark/geotrellis/layer.py
```diff
@@ -541,25 +541,34 @@ def cut_tiles(self, layer_metadata, resample_method=ResampleMethod.NEAREST_NEIGH
...
 
-        if isinstance(layer_metadata, Metadata):
-            layer_metadata = layer_metadata.to_dict()
+        if isinstance(layout, Metadata):
+            layer_metadata = layout.to_dict()
+            srdd = self.srdd.tileToLayout(json.dumps(layer_metadata), ResampleMethod(resample_method))
+        elif isinstance(layout, TiledRasterLayer):
+            layer_metadata = layout.layer_metadata
+            srdd = self.srdd.tileToLayout(json.dumps(layer_metadata), ResampleMethod(resample_method))
+        elif isinstance(layout, GlobalLayout) or isinstance(layout, LocalLayout):
```
jbouffard: This isn't an issue just a style suggestion, but if you're testing if something is an instances of more than one type you can do:

```python
elif isinstance(layout, (LayoutLevel, GlobalLayout))
```

And that would be the equivalent of testing with the `or`.

geopyspark/geotrellis/layer.py
```diff
@@ -541,25 +541,34 @@ def cut_tiles(self, layer_metadata, resample_method=ResampleMethod.NEAREST_NEIGH
...
+            srdd = self.srdd.tileToLayout(layout, resample_method)
```
jbouffard: Should be `ResampleMethod(resample_method)`

PR #391, opened 2017-07-24
Cleanup and Renames

jbouffard:
This PR renames the Scala classes and objects so that they match the patten of their Python counterparts (ie. their names now end in `Layer`). The `cut_tiles` method is also removed.


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterRDD.scala
```diff
@@ -46,10 +46,10 @@ import java.util.ArrayList
 import scala.reflect._
 import scala.collection.JavaConverters._
 
-class SpatialTiledRasterRDD(
+class SpatialTiledRasterLayer(
```
echeipesh: File name wasn't changed to match class name

echeipesh:
Looks good after the left over file rename is fixed.

PR #393, opened 2017-07-24
LocalLayout is Now Constructed From tile_cols and tile_rows

jbouffard:
This PR makes `LocalLayout` except either a `tile_size`, `tile_cols`, or `tile_rows` as constructors. This'll give the user more control over how the `LocalLayout` is defined.

### DIFF
--- a/geopyspark/geotrellis/__init__.py
+++ b/geopyspark/geotrellis/__init__.py
@@ -258,20 +258,38 @@ Returns:
 GlobalLayout.__new__.__defaults__ = (256, None, 0.1)
 
 
-LocalLayout = namedtuple("LayoutDefinition", 'tile_size')
-"""TileLayout type that snaps the layer extent.
+class LocalLayout(namedtuple("LocalLayout", 'tile_cols tile_rows')):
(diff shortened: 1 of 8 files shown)


geopyspark/geotrellis/__init__.py
```diff
@@ -258,20 +258,38 @@ def _asdict(self):
 GlobalLayout.__new__.__defaults__ = (256, None, 0.1)
 
 
-LocalLayout = namedtuple("LayoutDefinition", 'tile_size')
```
echeipesh: My concern here is that exposing cols/rows is inconstant with `GlobalLayout` which can only receive tile size. These need to feel consistent and square tile in both cases is most common case.

jbouffard: `LocalLayout` can still be constructed from `tile_size` https://github.com/jbouffard/geopyspark/blob/feature/layout-type-local-layout/geopyspark/geotrellis/__init__.py#L288 It just won't have it as a property.

echeipesh: Oh yeah, missed that, that's good.

PR #397, opened 2017-07-25
reproject and tile_to_layout Method Updates

jbouffard:
This Pr updates the `reproject` and `tile_to_layout` methods for `RasterLayer` and `TiledRasterLayer`. This resolves the API by having the buffered reproject being used in `tile_to_layout` instead of `reproject` which will not sample past the rasters boundaries.


geopyspark/geotrellis/layer.py
```diff
@@ -474,39 +472,31 @@ def collect_metadata(self, layout=LocalLayout()):
...
-            srdd = self.srdd.reproject(target_crs, resample_method)
-            return RasterLayer(self.pysc, self.layer_type, srdd)
-        else:
-            return _reproject(target_crs, layout, resample_method, self)
+        srdd = self.srdd.reproject(target_crs, resample_method)
 
-    def tile_to_layout(self, layout=LocalLayout(), resample_method=ResampleMethod.NEAREST_NEIGHBOR):
+        return RasterLayer(self.pysc, self.layer_type, srdd)
+
+    def tile_to_layout(self, layout=LocalLayout(), target_crs=None, resample_method=ResampleMethod.NEAREST_NEIGHBOR):
```
CloudNiner: 👍 I like `target_crs` here over just `crs`, its helpful in clearing up the intent of the parameter.

PR #408, opened 2017-07-26
Preserve Zoom Level

jbouffard:
This PR allows for the preservation of the zoom level of a `TiledRasterLayer` when going to a numpy RDD and back. In addition, a check has been added to the `write` function that makes sure the given layer has a `zoom_level` before trying to write it.

### DIFF
--- a/geopyspark/geotrellis/catalog.py
+++ b/geopyspark/geotrellis/catalog.py
@@ -12,7 +12,7 @@ import shapely.wkb
 from geopyspark import map_key_input, get_spark_context
 from geopyspark.geotrellis.constants import LayerType, IndexingMethod
 from geopyspark.geotrellis.protobufcodecs import multibandtile_decoder
-from geopyspark.geotrellis import Metadata, Extent, deprecated
+from geopyspark.geotrellis import Metadata, Extent, deprecated, Log
 from geopyspark.geotrellis.layer import TiledRasterLayer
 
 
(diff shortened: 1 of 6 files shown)


geopyspark/geotrellis/catalog.py
```diff
@@ -423,19 +423,22 @@ def write(uri,
             be in camel case. If both options and keywords are set, then the options will be used.
     """
 
+    if not tiled_raster_layer.zoom_level:
+        raise ValueError("The given layer does not have a zoom_level", tiled_raster_layer)
```
echeipesh: This is a problem. In GeoTrellis when we're writing a floating layer it gets the  zoom level of 0. In PySpark the `LocalLayers` layers have `zoom_level` of `None`.

Since we must be able to write a `LocalLayer`, we should use zoom level of 0. On the read side we when not given a zoom we should read the 0th level but report the `zoom_level` as `None`. I think this will keep a consistent view of what is going on.

jbouffard: Just so I understand, you're saying that if a `zoom` is not given to `query` then the resulting `TiledRasterLayer` should have a `zoom_level` of `None`, right? I think that's a bit confusing from an API standpoint. If I were doing a read I'd expect the resulting `TiledRasterLayer` to have the `zoom_level` I gave it. In the case that I didn't supply a `zoom`, I'd expect something back with a `zoom_level` of 0. Plus, it's a bit odd that `query("spatial", my_uri, layer_name, layer_zoom=0)` and `query("spatial", my_uri, layer_name)` would return different `TiledRasterLayer`s even though they both read from the same catalog and zoom.

PR #421, opened 2017-07-28
datetime for Temporal Parameters

### LINKED ISSUE #420: Support More Types for the instant Parameter/Attribute in Spatial-Temporal Objects/Functions
As of right now, the `instant` parameter or attribute for any function or object is an `int`. However, this is vague and hard to work with. We should support more types that can be used for instance to make it easy to work with spatial-temporal data in GeoPySpark.

jbouffard:
This PR makes it so that all parameters that are time components are now expected to be `datetime.datetime` instances. This'll make it easier to work with spatial-temporal data in GeoPySpark

This PR resolves #420

### DIFF
--- a/geopyspark/geotrellis/__init__.py
+++ b/geopyspark/geotrellis/__init__.py
@@ -9,13 +9,6 @@ from geopyspark import get_spark_context
 from geopyspark.geotrellis.constants import CellType, NO_DATA_INT
 
 
-_EPOCH = datetime.datetime.utcfromtimestamp(0)
-
-
-def _convert_to_unix_time(date_time):
-    return int((date_time - _EPOCH).total_seconds() * 1000)
-
-
 def deprecated(func):
     """This is a decorator which can be used to mark functions
     as deprecated. It will result in a warning being emmitted
@@ -235,21 +228,15 @@ class TemporalProjectedExtent(namedtuple("TemporalProjectedExtent", 'extent inst
 
     Args:
         extent (:class:`~geopyspark.geotrellis.Extent`): The area the raster represents.
-        instant (``datetime.datetime``): The time stamp of the raster. This is
-            an instance of ``datetime.datetime`` that references the correct date and time.
+        instant (``datetime.datetime``): The time stamp of the raster.
         epsg (int, optional): The EPSG code of the CRS.
         proj4 (str, optional): The Proj.4 string representation of the CRS.
 
     Attributes:
         extent (:class:`~geopyspark.geotrellis.Extent`): The area the raster represents.
-        instant (int): The time stamp of the raster. This is an ``int`` which represents the
-            timestamp in miliseconds.
+        instant (``datetime.datetime``): The time stamp of the raster.
         epsg (int, optional): The EPSG code of the CRS.
         proj4 (str, optional): The Proj.4 string representation of the CRS.
-
-    Note:
-        The ``instant`` attribute is in milliseconds since the 1970 epoch. Thus, it may be
-        different than the timestamp given by the ``datetime.datetime`` instance.
 
     Note:
         Either ``epsg`` or ``proj4`` must be defined.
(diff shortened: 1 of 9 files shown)


geopyspark/geotrellis/__init__.py
```diff
@@ -227,23 +235,32 @@ class TemporalProjectedExtent(namedtuple("TemporalProjectedExtent", 'extent inst
 
     Args:
         extent (:class:`~geopyspark.geotrellis.Extent`): The area the raster represents.
-        instance (int): The time stamp of the raster.
+        instant (``datetime.datetime``): The time stamp of the raster. This is
+            an instance of ``datetime.datetime`` that references the correct date and time.
```
echeipesh: The second line is a little redundant, the signature already gives the instance. Pointing out because shorter docs are more readable.

geopyspark/geotrellis/__init__.py
```diff
@@ -227,23 +235,32 @@ class TemporalProjectedExtent(namedtuple("TemporalProjectedExtent", 'extent inst
...
+
+    Note:
         Either ``epsg`` or ``proj4`` must be defined.
     """
 
     __slots__ = []
 
     def __new__(cls, extent, instant, epsg=None, proj4=None):
+        if isinstance(instant, datetime.datetime):
+            instant = _convert_to_unix_time(instant)
```
echeipesh: Should this conversion happen in ProtoBuf decoders ? That way the python meaningful `datetime` instance can hang around as long as possible.

geopyspark/geotrellis/catalog.py
```diff
@@ -325,10 +329,10 @@ def query(layer_type,
                 All other types are restricted to ``Polygon`` and ``MulitPolygon``.
 
             If not specified, then the entire layer will be read.
-        time_intervals (list, optional): A list of strings that time intervals to query.
-            The strings must be in a valid date-time format. This parameter is only used when
-            querying spatial-temporal data. The default value is, None. If None, then only the
-            spatial area will be querried.
+        time_intervals (list, optional): A list of ``datetime.datetime`` instanaces that represent
```
echeipesh: Is there a convention for giving the type of objects expected in the list in the signature rather than spelling it out?

geopyspark/geotrellis/__init__.py
jbouffard: Yeah, that'd be better placed in the `protobufcodecs` module.

geopyspark/geotrellis/catalog.py
jbouffard: Honestly, I've never seen a convention around a `list` parameter. Though, I do think it'd be better if it was `[datetime.datetime]`.

geopyspark/geotrellis/catalog.py
```diff
@@ -415,8 +423,9 @@ def write(uri,
             method used to orginize the saved data. Depending on the type of data within the layer,
             only certain methods are available. Can either be a string or a ``IndexingMethod``
             attribute.  The default method used is, ``IndexingMethod.ZORDER``.
-        time_unit (str, optional): Which time unit should be used when saving spatial-temporal data.
-            While this is set to None as default, it must be set if saving spatial-temporal data.
+        time_unit (str or :class:`~geopyspark.geotrellis.constants.TimeUnit`, optional): Which time
+            unit should be used when saving spatial-temporal data.  While this is set to ``None``
+            as default, it must be set if saving spatial-temporal data.
```
echeipesh: Noticed that this doc string is misleading. This controls temporal resolution of the index. Approximately at what time interval the records are spaced temporally. Is there an image for every day, hour, minute?

geopyspark/geotrellis/catalog.py
echeipesh: That makes sense, lets adopt it?

jbouffard: I think we should. I know we've already used in some other docstrings.

PR #424, opened 2017-07-31
Supply TMS server API

jpolchlo:
To make use of the TMS server, it is necessary to be able to start up and shut down the service as well as to fetch the URI pattern for the resulting endpoint.  This PR supplies that interface.


jbouffard:
Looks like you're dealing with Travis memory problems https://travis-ci.org/locationtech-labs/geopyspark/jobs/259449011#L4002 I'd try restarting the build to see if it passes. If becomes a problem, we'll have to look into updating the tests again.

geopyspark/geotrellis/tms.py
```diff
@@ -115,18 +116,86 @@ class TMS(object):
...
+                self.server.bind(host, requested_port)
+            else:
+                self.server.bind(host)
+            self.bound = True
+            self._port = self.server.port()
+            self._host = [l for l in
+                          ([ip for ip in socket.gethostbyname_ex(socket.gethostname())[2] if not ip.startswith("127.")][:1],
+                           [[(s.connect(('8.8.8.8', 53)), s.getsockname()[0], s.close()) for s in
+                             [socket.socket(socket.AF_INET, socket.SOCK_DGRAM)]][0][1]])
+                          if l][0][0]
```
echeipesh: What is happening here? What the `host` not be just a passthrough from the parameter ?

jpolchlo: No.  If we just pass through "0.0.0.0", and you request the URL pattern, you can't just copy that into `geojson.io`.  This grabs the system's IP, so that it's a meaningful address.

echeipesh: Hmm, that makes sense for `0.0.0.0` case, should that be the only thing that triggers this behavior? Otherwise the host name must have been specific enough for Akka bind to actually find the interface and bind to some port already.

jpolchlo: I'm open to suggestions.  Not 100% sure what the likely range of host arguments will be.

lossyrob: You can't copy `http://0.0.0.0:xx/...` into geojson.io? I know you can with `localhost`

jpolchlo: Is the right thing to do to allow host to be unspecified?  In those cases, we can look up the local IP.  In all other cases, we pass through the supplied host?  Perhaps the global IP URI can be explicitly grabbed in cases where it's needed?  Not at all sure of what constitutes best practice here.

echeipesh: This behavior I think is the least confusing:
- `host` is unspecified => bind to `localhost`
   this is the safest choice
- `host` is `0.0.0.0` => perform reverse lookup
   0.0.0.0 implies that this this service is supposed to be accessible from outside, so lets do our best.
- `host` is a hostname => bind and report it as given
   Presumably they gave a hostname that somehow binds to an interface, so it should work on those terms.

AFAIK `0.0.0.0` is the only case where we break hostname, interface  1-1 mapping and have to do something special to get a useful URI.

PR #425, opened 2017-07-31
TiledRasterLayer.tile_to_layout Bug Fix

jbouffard:
This PR fixes a bug where the `TiledRasterLayer.tile_to_layout` method would always return a new layer with a `zoom_level` of `None` regardless of the layout given. Now, the new layer will keep the given/produced `zoom_level` when the layout is either a `LocalLayout` or `GlobalLayout`.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
@@ -123,6 +123,7 @@ class TemporalTiledRasterLayer(
 
   def tileToLayout(
     layoutDefinition: LayoutDefinition,
+    zoom: Option[Int],
     resampleMethod: ResampleMethod
   ): TiledRasterLayer[SpaceTimeKey] = {
     val mapKeyTransform =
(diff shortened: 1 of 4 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
```diff
@@ -123,6 +123,7 @@ class TemporalTiledRasterLayer(
 
   def tileToLayout(
     layoutDefinition: LayoutDefinition,
+    zoom: Option[Int],
     resampleMethod: ResampleMethod
   ): TiledRasterLayer[SpaceTimeKey] = {
```
echeipesh: Is this call to `.tile_to_layout` uncovered now ? https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/layer.py#L997

jbouffard: I covered it here https://github.com/jbouffard/geopyspark/blob/bug-fix/tiledlayer-tile-to-layout/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala#L95

echeipesh: 👍

PR #426, opened 2017-08-01
AttributeStore

echeipesh:
Exposes AttributeStore as an object in geopyspark.
Because it is so central to many GeoTrellis operations it was given a class rather than just read/write methods.

Because we can't really keep track of what type the objects are, all reading and writing is done on `dict` values.

Motivating use case is saving and reading histograms:
```py
In [1]: import geopyspark as gps

In [2]: store = gps.geotrellis.catalog.AttributeStore("s3://azavea-datahub/catalog")

In [3]: store.layer("us-pennsylvania-philadelphia-nlcd2011-landcover-30m-epsg3857",7).read("histogram")
Out[3]:
[[95, 43],
 [90, 133],
 [82, 14],
 [81, 51],
 [71, 5],
 [52, 33],
 [43, 31],
 [42, 2],
 [41, 394],
 [31, 12],
 [24, 623],
 [23, 952],
 [22, 892],
 [21, 979],
 [11, 188]]

In [4]: gps.Histogram.from_dict(Out[3])
Out[4]: <geopyspark.geotrellis.histogram.Histogram at 0x109ffd400>

In [5]: gps.Histogram.from_dict(Out[3]).min_max()
Out[5]: (11, 95)
```

### DIFF
--- a/geopyspark/geotrellis/catalog.py
+++ b/geopyspark/geotrellis/catalog.py
@@ -4,7 +4,6 @@
 import json
 from collections import namedtuple
 from urllib.parse import urlparse
-from itertools import groupby
 from shapely.geometry import Polygon, MultiPolygon, Point
 from shapely.wkt import dumps
 import shapely.wkb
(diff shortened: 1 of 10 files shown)


geopyspark/geotrellis/catalog.py
```diff
@@ -4,19 +4,19 @@
...
-from geopyspark import map_key_input, get_spark_context
+from geopyspark import map_key_input, get_spark_context, scala_companion
 from geopyspark.geotrellis.constants import LayerType, IndexingMethod, TimeUnit
 from geopyspark.geotrellis.protobufcodecs import multibandtile_decoder
 from geopyspark.geotrellis import Metadata, Extent, deprecated, Log
 from geopyspark.geotrellis.layer import TiledRasterLayer
 
 
-__all__ = ["read_layer_metadata", "get_layer_ids", "read_value", "query", "write"]
+__all__ = ["read_layer_metadata", "get_layer_ids", "read_value", "query", "write", "AttributeStore"]
```
jbouffard: Is `AttributeStore` something the users will be working with in the future? Because I'm not sure if it makes sense to add it to `__all__` if that's not the case.

geopyspark/geotrellis/catalog.py
```diff
@@ -453,3 +458,94 @@ def write(uri,
                                     tiled_raster_layer.srdd,
                                     TimeUnit(time_unit).value,
                                     IndexingMethod(index_strategy).value)
+
+
+class Attributes(object):
```
jbouffard: Will `Attributes` ever be seen or used by the user? If not, it might be good to add a `_` at the beginning so that people know it's a backend class.

jbouffard: If we're not going to make it private, then this class will need docstrings.

geopyspark/geotrellis/catalog.py
```diff
@@ -453,3 +458,94 @@ def write(uri,
...
+        return self.write(name, value)
+
+    def __delitem__(self, name):
+        return self.delete(name)
+
+    def read(self, name):
+        """Read layer attribute by name as a dict
+
+        Args:
+           name (str) Attribute name
```
jbouffard: These docstrings need to be finished (needs a Return).

geopyspark/geotrellis/catalog.py
```diff
@@ -453,3 +458,94 @@ def write(uri,
...
+    def delete(self, name):
+        """Delete attribute by name
+
+        Args:
+            name (str): Attribute name
+        """
+        self.store.underlying.delete(self.layer_id, name)
+
+
+class AttributeStore(object):
```
jbouffard: Same question as the one on `Attributes`.

geopyspark/geotrellis/catalog.py
```diff
@@ -453,3 +458,94 @@ def write(uri,
...
+        self.underlying = scala_companion("geotrellis.spark.io.AttributeStore").apply(uri)
+
+    def layer(self, name, zoom=None):
+        """Layer Attributes object for given layer
+        Args:
+            name (str): Layer name
+            zoom (int, optional): Layer zoom
+
+        Returns:
+            :class:`~geopyspark.geotrellis.catalog.Attributes
```
jbouffard: Missing the closing `

geopyspark/geotrellis/catalog.py
```diff
@@ -453,3 +458,94 @@ def write(uri,
...
+        """
+        return Attributes(self, name, zoom)
+
+    def layers(self):
+        """List all layers Attributes objects"""
```
jbouffard: Docstrings incomplete.

geopyspark/tests/io_tests/s3_geotiff_rdd_test.py
```diff
@@ -67,26 +68,36 @@ def read_multiband_geotrellis(self, opt=options):
...
+    def test_segment_tiles(self):
+        # GeoTrellis will read GeoTiff Segments given no window size
+        # Retile them to match Rasterio read and check the cell values
+        geotrellis_tiles = self.read_multiband_geotrellis().tile_to_layout(LocalLayout(512))
+        geotrellis_tiles = [tile[1] for tile in geotrellis_tiles.to_numpy_rdd().collect()]
         rasterio_tiles = self.read_geotiff_rasterio([self.file_path], False)
 
         for x, y in zip(geotrellis_tiles, rasterio_tiles):
-            self.assertTrue((x.cells == y['cells']).all())
+            print('\n')
```
jbouffard: These `print` statements should probably be removed.

geopyspark/geotrellis/histogram.py
```diff
@@ -24,6 +26,14 @@ class Histogram(object):
     def __init__(self, scala_histogram):
         self.scala_histogram = scala_histogram
 
+    @classmethod
+    def from_dict(cls, d):
```
jbouffard: Sorry to nitpick, but `d` should probably be renamed to something else if this method will be used by users. Also, needs the remaining docstrings.

geopyspark/geotrellis/histogram.py
```diff
@@ -174,3 +184,10 @@ def merge(self, other_histogram):
         """
 
         return Histogram(self.scala_histogram.merge(other_histogram.scala_histogram))
+
+    def to_dict(self):
+        """Encodes histogram as a dictionary"""
```
jbouffard: Docstrings

geopyspark/tests/io_tests/s3_geotiff_rdd_test.py
```diff
@@ -67,26 +68,36 @@ def read_multiband_geotrellis(self, opt=options):
...
     def test_windowed_tiles(self):
         geotrellis_tiles = self.read_multiband_geotrellis({"s3Client": "mock", "maxTileSize": 256})
+        geotrellis_tiles = [tile[1] for tile in geotrellis_tiles.to_numpy_rdd().collect()]
         rasterio_tiles = self.read_geotiff_rasterio([self.file_path], True)
 
         self.windowed_result_checker(geotrellis_tiles)
 
         for x, y in zip(geotrellis_tiles, rasterio_tiles):
-            self.assertTrue((x.cells == y['cells']).all())
+            print('\n')
```
jbouffard: Same as above.

setup.cfg
```diff
@@ -1,2 +1,6 @@
 [metadata]
 license_file = LICENSE
+
+[flake8]
```
jbouffard: Are we adding flake8 testing as well?

geopyspark/geotrellis/catalog.py
echeipesh: Yes, its its something that they'll be making and holding around. A lot of the workflows don't make a lot of sense without it.

geopyspark/geotrellis/catalog.py
echeipesh: This could be private. Do you think its cleaner to make it `_Attributes` or an inner class to `AttributeStore` ? Its basically impossible to have an `Attributes` without an instance of `AttributeStore` so inner class may be appropriate?

setup.cfg
echeipesh: This is something that I'm using for local linting, up for debate if these settings belong in the repo.  Doesn't seem like it would hurt to have it configured for people that use it.

geopyspark/geotrellis/catalog.py
jbouffard: Actually, I'm not sure if this would work as being private. Since you need to access the `read`, `write`, and `delete` methods, it seems odd to be working with a class that's supposed to be hidden. I guess it makes sense then to have it be inside the `AttributeStore` since what we're really controlling is the creation of `Attributes`.

If we don't want the user to have any access to `Attributes` then we'd have to move some of `Attributes`'s methods over to `AttributeStore` and have some kind caching where we store `(layer_name, zoom) -> Attributes` in the `AttributeStore` instance.

setup.cfg
jbouffard: I see. I'm not familiar with flake8, but with pylint we had to ignore a lot of errors and warnings https://github.com/locationtech-labs/geopyspark/blob/master/pylintrc#L7 So anything we ignore for pylint will also need to be ignored for flake8 as well.

geopyspark/tests/io_tests/s3_geotiff_rdd_test.py
echeipesh: They were really helpful when things went wrong and are otherwise hidden, I'd actually like to save them for when/if this throws again.

jbouffard: That makes sense. I'm fine with them staying then.

PR #430, opened 2017-08-01
Refactored the saveStitched Method

jbouffard:
This PR refactors the `saveStitched` method so that it is easier to use. This was done mainly by changing one of the parameters to be an `Extent`, simplifying the code, and renaming certain methods.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
@@ -192,36 +192,36 @@ class SpatialTiledRasterLayer(
     PythonTranslator.toPython[Tile, ProtoTile](contextRDD.stitch.tile)
   }
 
-  def save_stitched(path: String): Unit =
-    _save_stitched(path, None, None)
+  def saveStitched(path: String): Unit =
+    saveStitched(path, None, None)
 
-  def save_stitched(path: String, cropBounds: ArrayList[Double]): Unit =
-    _save_stitched(path, Some(cropBounds), None)
+  def saveStitched(path: String, cropBounds: java.util.Map[String, Double]): Unit =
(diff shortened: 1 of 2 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
```diff
@@ -192,36 +192,36 @@ class SpatialTiledRasterLayer(
...
+
+  def saveStitched(path: String, cropBounds: java.util.Map[String, Double]): Unit =
+    saveStitched(path, Some(cropBounds), None)
+
+  def saveStitched(
+    path: String,
+    cropBounds: java.util.Map[String, Double],
+    cropDimensions: ArrayList[Int]
+  ): Unit =
+    saveStitched(path, Some(cropBounds), Some(cropDimensions))
```
echeipesh: It would be safer to use `Option(cropBounds)` here, that will turn the value to `None` if its passed a `null`, which happens if its given a python `None`.

jbouffard: Do you still think that's needed even with the checks done on the python side? https://github.com/jbouffard/geopyspark/blob/d6f4a0f689a792dd049d8444dbea6f62a72eca7e/geopyspark/geotrellis/layer.py#L1133

echeipesh: Was mostly pointing out a pattern that I think is good practice. Its definitely guarded in this case but  potentially this may have another call origin that won't do the same check or some other tomfoolery.

jbouffard: Ah, okay. I'll update the methods then so that they provide that extra protection.

PR #436, opened 2017-08-02
query and read_value Fix For Spatial-Temporal Data

jbouffard:
This PR fixes a bug present in the `query` and `read_value` functions. This bug appeared when the user tries to work with spatial-temporal data. The cause of this issue was that the ISO strings sent from Python were not properly formatted.

### DIFF
--- a/geopyspark/geotrellis/catalog.py
+++ b/geopyspark/geotrellis/catalog.py
@@ -11,7 +11,7 @@ import shapely.wkb
 from geopyspark import map_key_input, get_spark_context, scala_companion
 from geopyspark.geotrellis.constants import LayerType, IndexingMethod, TimeUnit
 from geopyspark.geotrellis.protobufcodecs import multibandtile_decoder
-from geopyspark.geotrellis import Metadata, Extent, deprecated, Log
+from geopyspark.geotrellis import Metadata, Extent, deprecated, Log, _convert_to_unix_time
 from geopyspark.geotrellis.layer import TiledRasterLayer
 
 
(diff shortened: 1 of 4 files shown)


geopyspark/geotrellis/catalog.py
```diff
@@ -365,14 +369,14 @@ def query(layer_type,
 
     else:
         if time_intervals:
-            time_intervals = [time.isoformat() for time in time_intervals]
+            time_intervals = [time.isoformat() + "Z" for time in time_intervals]
```
echeipesh: Thats not safe way to convert the datetime. `Z` prefix indicates that the timestamp is in UTC timezone, which it may or may not be.

```py
In [38]: dt = datetime.combine(d, t)

In [39]: d = date(2005, 7, 14)

In [40]: dt = datetime.combine(d, t)

In [41]: dt.isoformat()
Out[41]: '2005-07-14T12:30:00'

In [42]: dt.astimezone(pytz.utc).isoformat()
Out[42]: '2005-07-14T16:30:00+00:00'

In [43]: dt.astimezone(eastern).isoformat()
Out[43]: '2005-07-14T12:30:00-04:00'
```

`datetime` instance may already have timezone set. If its not there is an implied system default (EST in my case). Probably the safest way is to convert to UTC time zone. I believe the java parser will accept timestamp with `+00:00` for timezone just fine.

jbouffard: Good catch. I'm not sure how let that one slip past me. I'll put of a fix for it in a couple of minutes.

PR #440, opened 2017-08-04
Implement multi-value lookup on RDD

echeipesh:
There is an interesting behavior in GeoTrellis where `Pyramid` will erase the partitioner, despite its own best efforts. Until that is fixed the multi-value lookup is going to fall back to filtering for all but the base level of the pyramid.

Transition from filter fallback to multivalue lookup:
screen shot 2017-08-04 at 3 03 54 pm


echeipesh:
Requires https://github.com/locationtech/geotrellis/pull/2311 to fix the referenced issue

PR #442, opened 2017-08-05
Layer IO SPI Usage

echeipesh:
Uses newly available GeoTrellis layer IO SPI to construct the required classes

- [x] reimplement `write`
- [x] reimplement `query`
- [x] `query` loses `layer_type` argument
- [x] Add python `TileReader`
- [x] Remove existing factory methods


echeipesh:
Able to verify that reading/reprojecting/writing from UTM layer works in this branch and still displays problems on master even after https://github.com/locationtech-labs/geopyspark/pull/445

Visual v
screen shot 2017-08-08 at 12 28 53 am
erification:

The GeoJSON bounding box is the are we expect to cover by reprojecting UTM bounding box to LatLng.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerReaderWrapper.scala
```diff
@@ -0,0 +1,177 @@
...
+      Option(numPartitions).map(_.toInt).getOrElse {
+        val tileBytes = (layerMetadata.cellType.bytes
+          * layerMetadata.layout.tileLayout.tileCols
+          * layerMetadata.layout.tileLayout.tileRows)
+        // Aim for ~16MB per partition
+        val tilesPerPartition = (1 << 24) / tileBytes
+        // TODO: consider temporal dimension size as well
+        val expectedTileCount: Long = layerQuery(layerMetadata).map(_.toGridBounds.sizeLong).reduce( _ + _)
+        math.max(1, (expectedTileCount / tilesPerPartition).toInt)
+      }
```
echeipesh: @lossyrob This is the part you're concerned about

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/AttributeStoreWrapper.scala
```diff
@@ -0,0 +1,70 @@
...
+  def write(layerName: String, zoom: Int, attributeName: String, value: String): Unit = {
+    val id = LayerId(layerName, zoom)
+    if (value == null) return
+    val json = value.parseJson // ensure we actually have JSON here
+    attributeStore.write(id, attributeName, json)
+  }
+
+  def delete(layerName: String, zoom: Int, name: String): Unit = {
+    val id = LayerId(layerName, zoom)
+    attributeStore.delete(id, name)
```
jbouffard: Should there be a check to make sure that the layer being deleted actually exists?

geopyspark/geotrellis/catalog.py
```diff
@@ -16,199 +17,34 @@
...
-    else:
-        return True
-
-
-def read_layer_metadata(layer_type,
-                        uri,
+def read_layer_metadata(uri,
                         layer_name,
-                        layer_zoom,
-                        options=None,
```
echeipesh: They're part of the `uri` structure now: http://geotrellis.readthedocs.io/en/latest/guide/tile-backends.html?highlight=cassandra#backend-uri

geopyspark/geotrellis/catalog.py
```diff
@@ -253,53 +83,68 @@ def read_value(layer_type,
...
     else:
-        options = options or kwargs or {}
+        store = AttributeStore.build(uri)
+        _cached_stores[uri] = store
 
-        if zdt:
-            zdt = zdt.astimezone(pytz.utc).isoformat()
-        else:
-            zdt = ''
+    reader = ValueReader(uri, layer_name, layer_zoom, store)
```
echeipesh: I am not sure, but I don't think so. The main cost we need to cache is the layer metadata lookup. We need to avoid fetching it on every tile request. Having an instance of `AttributeStoreWrapper` should take care of that. 

All the remaining tile readers are basically just interface class creations leading to `.read` call. I could see the python caching machinery taking longer to work then the actual JVM class creation. So I decided to put this question off until we're benchmarking the tile server.

geopyspark/geotrellis/catalog.py
```diff
@@ -253,53 +83,68 @@ def read_value(layer_type,
...
+    return reader.read(col, row, zdt)
 
-        if uri not in _mapped_cached:
-            _construct_catalog(get_spark_context(), uri, options)
 
-        cached = _mapped_cached[uri]
+class ValueReader(object):
```
echeipesh: Thats up for discussion. The really good thing about `ValueReader` being a class is that it holds an exact instance of `AttributeStore`. So if something goes wrong its a very direct link to what is likely causing a failure. 

It feels like this should be an option in the API that can expose more functionality aside from `read`. Like this would be the place to implement a way to `read_extent` and `read_thumbnail` and all the other functions that basically use Collections API.

It breaks from "everything is a function" we've strived for in the API but `AttributeStore` did that and actually ended up much more usable, this could be another case.

geopyspark/geotrellis/catalog.py
```diff
@@ -253,53 +83,68 @@ def read_value(layer_type,
...
+                Defaults to ``self.zoom``
 
-def query(layer_type,
-          uri,
+        Returns:
+            :class:`~geopyspark.geotrellis.Tile`
+        """
+
+        zoom = zoom or self.zoom or 0
+        zoom = zoom and int(zoom)
```
echeipesh: Maybe too fancy, if `zoom` is an instance of `str` it will pass line 133 and it would be converted here.

geopyspark/geotrellis/catalog.py
```diff
@@ -340,84 +185,53 @@ def query(layer_type,
...
 
     Returns:
         :class:`~geopyspark.geotrellis.rdd.TiledRasterLayer`
     """
-
-    options = options or kwargs or {}
-    layer_zoom = layer_zoom or 0
-
     pysc = get_spark_context()
+    attribute_store = attribute_store or AttributeStore.build(uri)
```
echeipesh: Thats a good point, wrote this part before the cache

geopyspark/geotrellis/catalog.py
```diff
@@ -253,53 +83,68 @@ def read_value(layer_type,
...
+def query(uri,
           layer_name,
           layer_zoom=None,
           query_geom=None,
           time_intervals=None,
           query_proj=None,
-          options=None,
           num_partitions=None,
-          **kwargs):
+          attribute_store=None):
```
echeipesh: The idea is to have a way to pass in a specific store into the query function and bypass the creation or cache lookup (as you point out). What do you think?

geopyspark/geotrellis/catalog.py
```diff
@@ -340,84 +185,53 @@ def query(layer_type,
...
+    else:
+        raise TypeError("Could not query intersection", query_geom)
 
-    key = map_key_input(LayerType(layer_type).value, True)
+    if isinstance(query_proj, int):
+        query_proj = str(query_proj)
 
-    num_partitions = num_partitions or pysc.defaultMinPartitions
+    time_intervals = time_intervals or []
+    time_intervals = [time.astimezone(pytz.utc).isoformat()
```
echeipesh: Looks good to me.

geopyspark/geotrellis/catalog.py
```diff
@@ -475,19 +286,39 @@ class AttributeStore(object):
...
-        self.underlying = scala_companion("geotrellis.spark.io.AttributeStore").apply(uri)
+        pysc = get_spark_context()
+        try:
+            self.wrapper = pysc._gateway.jvm.geopyspark.geotrellis.io.AttributeStoreWrapper(uri)
+        except Py4JJavaError as err:
+            raise ValueError(err.java_exception.getMessage())
+
+
+    @classmethod
+    def build(cls, uri):
```
echeipesh: I think I'm out of the loop there, what do you mean ?

geopyspark/geotrellis/constants.py
```diff
@@ -27,6 +27,19 @@ class LayerType(Enum):
     """
     SPACETIME = 'spacetime'
 
+    @classmethod
```
echeipesh: Good point on both.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/AttributeStoreWrapper.scala
echeipesh: `.delete` on layer that doesn't exist is jus a no-op right now. Although it would be nice to return `True`/`False`.

geopyspark/geotrellis/catalog.py
jbouffard: I'm not against the idea, but what benefit would passing in a pre existing `AttributeStore` have? I'm just trying to wrap my head around the new API.

geopyspark/geotrellis/catalog.py
jbouffard: Sorry, I was wrong about the `__slots__`. However, I'm still not sure this method is needed. Why can't the logic in `build` be moved to `__init__`?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/AttributeStoreWrapper.scala
jbouffard: Yeah, I think it would be nice to let the user know if what they wanted deleted doesn't actually exist.

geopyspark/geotrellis/catalog.py
echeipesh: Its the same deal as `ColorMap.build`, the `__init__` can assign a attributes to a class but it can't choose to return a different instance. For that you have to either use `__new__` or a class method like `build`. The last refactor removed the use of it though (it was confusing what the way it was being used) so it can go away until it is needed.

geopyspark/geotrellis/catalog.py
echeipesh: `AttributeStore` and `LayerReader` must not always share the same storage mechanism, although thats true by default. You could be pulling tiles from S3 but have the attributes stored in Accumulo/PostgreSQL/filesystem. 

Being able to decouple the attribute store instance used and the layer URI is whats required to take advantage of that feature. Having an optional argument for it is minimal way to enable that use case.

geopyspark/geotrellis/catalog.py
jbouffard: I see. That makes sense. I think as long as the differences between `__init__` and `build` are clear in the docs then it'd fine.

geopyspark/geotrellis/catalog.py
jbouffard: Ah, okay. I think that implementation works then. I still think there needs to be a check for the `AttributeStore` in the cache if one is not given, but otherwise I'm convinced.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/AttributeStoreWrapper.scala
echeipesh: Checking into it unfortunately the underlying Scala function does not provide that information, so it would require read/delete pattern. Adding `AttributeStore.contains` for ease of writing that kind of check though.

PR #445, opened 2017-08-07
Fix TiledRasterLayer.reproject

echeipesh:
`TiledRasterLayer.rdd` is already tiled, tileToLayout does not need to be called before reprojection.


echeipesh:
This seems to be the culprit of the projection weirdness that I've been seeing when working with LiDAR raster. Follow up issue is why `tileToLayout` effected it so, at worst it should have been a wasted work.

PR #452, opened 2017-08-08
Core Concepts Docs for 0.2

jbouffard:
This PR adds the `core-concept` notebook that goes over the fundamental concepts used in GeoPySpark.

### DIFF
--- a/docs/notebook-demos/core-concepts.ipynb
+++ b/docs/notebook-demos/core-concepts.ipynb
@@ -1,0 +1,418 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import datetime\n",
+    "import numpy as np\n",
+    "import geopyspark as gps"
+   ]
+  },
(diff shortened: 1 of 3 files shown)


docs/notebook-demos/core-concepts.ipynb
```diff
@@ -0,0 +1,418 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "# Creates Metadata for a layer with rasters that have a cell type of int16 with the previously defined\n",
+    "# bounds, crs, extent, and layout definition.\n",
+    "gps.Metadata(bounds=bounds,\n",
```
echeipesh: `bounds` here is not defined, it needs to be assigned in the above section.

jbouffard: `bounds` is actually defined in the section preceding this one.

echeipesh: Hmm, weird, I copied and ran the workbook and the assignment was not there. I see it in the commit though.
Disregard this comment please.

PR #453, opened 2017-08-08
TMS: Multiband support, render tweak

lossyrob:
These changes update to use Multibands in the backend.
It also passes the tile into the python render function, instead of just the cells; the no_data_value can be useful to the render function.

### DIFF
--- a/geopyspark/geotrellis/tms.py
+++ b/geopyspark/geotrellis/tms.py
@@ -16,8 +16,8 @@ class TileRender(object):
     rendering functions.
 
     Args:
-        render_function (numpy.ndarray => bytes): A function to convert a numpy
-            array to a collection of bytes giving a binary image file.
+        render_function (Tile => bytes): A function to convert geopyspark.geotrellis.Tile
+            to a collection of bytes giving a binary image file.
 
     Attributes:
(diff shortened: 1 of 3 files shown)


geopyspark-backend/project/Version.scala
```diff
@@ -1,6 +1,6 @@
 object Version {
   val geopyspark = "0.2.0"
-  val geotrellis = "1.2.0-SNAPSHOT"
+  val geotrellis = "1.2.0-ROB-WORKSHOP"
```
jpolchlo: Looks like local configs got added by mistake.

lossyrob: dang

geopyspark/geotrellis/tms.py
```diff
@@ -16,8 +16,8 @@ class TileRender(object):
     rendering functions.
 
     Args:
-        render_function (numpy.ndarray => bytes): A function to convert a numpy
-            array to a collection of bytes giving a binary image file.
+        render_function (Tile => bytes): A function to convert geopyspark.geotrellis.Tile
```
jpolchlo: Just noticing this docstring is incorrect.  The render function should produce an Image, not bytes.

PR #454, opened 2017-08-09
Add overzooming capability to TMS servers

jpolchlo:
This PR allows TMS servers to overzoom or not based on the `allow_overzooming` flag passed to `TMS.build` (defaults to true).

This PR will be ready to review after confirming proper functioning in ...
 - [x] catalog-backed TMS servers
 - [x] RDD-backed TMS servers

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/tms/TileReader.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/tms/TileReader.scala
@@ -24,7 +24,7 @@ trait TileReader {
 
 object TileReaders {
 
-  private def rezoom(zoom: Int, x: Int, y: Int, maxZoom: Int, read: SpatialKey => Tile) = {
+  private def rezoom(zoom: Int, x: Int, y: Int, maxZoom: Int, read: SpatialKey => MultibandTile): MultibandTile = {
     val dz = zoom - maxZoom
     val key = SpatialKey((x / math.pow(2, dz)).toInt, (y / math.pow(2, dz)).toInt)
     val dx = x - key._1 * math.pow(2, dz).toInt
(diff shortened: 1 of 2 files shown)


jpolchlo:
I've tested this using S3 catalog layers (NLCD) on the Ktile vis server.  There are intermittent problems there—mostly having to do with odd `[errno 111] connection refused` errors from inside Ktile.  These problems show up after a few cycles of adding and removing TMS layers.  Will investigate.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/tms/TileReader.scala
```diff
@@ -24,18 +24,46 @@ trait TileReader {
...
+    private val maxZoom = zoomLevels.max
 
     def retrieve(zoom: Int, x: Int, y: Int) = {
-      val key = SpatialKey(x, y)
       Future {
-        val reader = layers.getOrElseUpdate(zoom, valueReader.reader[SpatialKey, Tile](LayerId(catalog, zoom)))
-        Try(reader(key)).toOption
+        if (overzooming && zoom > maxZoom) {
+          val reader = layers.getOrElseUpdate(maxZoom, valueReader.reader[SpatialKey, Tile](LayerId(catalog, maxZoom)))
+          Try(rezoom(zoom, x, y, maxZoom, reader(_))).toOption
```
echeipesh: Heads up, this `Try` is dangerous for debugging. You will want to `match` here for `Failure(e: ValueNotFoundError) => None`  and re-throw anything else.

jpolchlo:
A note: The Geotrellis vis server in GeoNotebook reaches into the guts of the tms server and calls the functions on the Scala class directly.  This can leave this object in an inconsistent state.  The current shape of the interface looks forward to the move to the Ktile server (and to direct use via geojson.io and the like).  Given the impending transition, it's not clear that this PR should trigger an update to the current version of GeoNotebook.  Until this PR is accompanied by an update to the GeoNotebook interface, odd behavior may arise.

PR #459, opened 2017-08-09
Sentinel Ingest Example Docs for 0.2

jbouffard:
This PR adds a notebook that contains an example on how to ingest Sentinel-2 images with GeoPySpark.

This PR is based on another, open PR #457

PR #460, opened 2017-08-09
Ingesting an Image Docs for 0.2

jbouffard:
This PR adds a notebook that contains an example on how to ingest an image with GeoPySpark.

This PR is based on another, open PR #459

### DIFF
--- a/docs/notebook-demos/ingesting-an-image.ipynb
+++ b/docs/notebook-demos/ingesting-an-image.ipynb
@@ -1,0 +1,207 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Ingesting an Image\n",
+    "\n",
+    "This example shows how to ingest a grayscale image and save the results locally. It is assumed that you have already read through the documentation on GeoPySpark before beginning this tutorial."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


docs/notebook-demos/ingesting-an-image.ipynb
```diff
@@ -0,0 +1,207 @@
...
+    "!curl -o /tmp/cropped.tif https://s3.amazonaws.com/geopyspark-test/example-files/cropped.tif"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## What is an Ingest?\n",
+    "\n",
+    "Before continuing on, it would be best to discuss what an ingest actually is. In the context of GeoPySpark, it means to read in data and format it in such a way so that it can be analyzed and saved for later usage. Thus, `TiledRasterLayer` can be thought of as this final end stage."
```
jpolchlo: This is extremely vague.

_Before continuing on, it would be best to briefly discuss what an ingest actually is.  When data are acquired, they may cover an arbitrary spatial extent in an arbitrary projection.  These data need to be regularized to some expected layout and cut into tiles.  After this step, we will possess a `TiledRasterLayer` that can be analyzed and saved for later use._

Perhaps a reference to the layer guide would be useful here?

jbouffard: Yeah, after rereading that section a link to the layers guide would be helpful.

PR #457, opened 2017-08-09
Layers Docs for 0.2

jbouffard:
This PR adds a notebook that contains documentation on the `RasterLayer` and `TiledRasterLayer` classes.

PR #467, opened 2017-08-14
Map Algebra Docs for 0.2

jbouffard:
This PR adds a notebook that goes over the map algebra operations.

This PR is based on another, open PR #457

### DIFF
--- a/docs/notebook-demos/map-algebra.ipynb
+++ b/docs/notebook-demos/map-algebra.ipynb
@@ -56,27 +56,22 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Map Algebra"
+    "# Map Algebra\n",
+    "\n",
+    "Given a set of raster layers, it may be desirable to combine and filter the content of those layers.  This is the function of _map algebra_.  Two classes of map algebra operations are provided by GeoPySpark: _local_ and _focal_ operations.  Local operations individually consider the pixels or cells of one or more rasters, applying a function to the corresponding cell values.  For example, adding two rasters' pixel values to form a new layer is a local operation.\n",
+    "\n",
+    "Focal operations consider a region around each pixel of an input raster and apply an operation to each region.  The result of that operation is stored in the corresponding pixel of the output raster.  For example, one might weight a 5x5 region centered at a pixel according to a 2d Gaussian to effect a blurring of the input raster.  One might consider this roughly equivalent to a 2d convolution operation.\n",
+    "\n",
+    "**Note:** Map algebra operations work only on `TiledRasterLayer`s, and if a local operation requires multiple inputs, those inputs must have the same layout and projection."
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Opertaions\n",
+    "## Local Operations\n",
     "\n",
-    "As of right now, two different types of operations can be performed on `TiledRasterLayer`s these being local and focal operations."
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "### Local Operations\n",
-    "\n",
-    "Local operations on `TiledRasterLayer`s can use `int`s, `float`s, or other `TiledRasterLayer`s.\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


docs/notebook-demos/map-algebra.ipynb
```diff
@@ -0,0 +1,655 @@
...
+    "tiled_layer.polygonal_min(geometry=poly_max, data_type=int)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Cost Distance\n",
+    "\n",
+    "`cost_distance` is a function that will compute the accumulative cost to the nearest geometry of each cell in a friction layer. Cells that fall within the given geometry will have an accumulative cost of 0.0, while cells that contain noData values will continue to be represented with noData. If desired, a `max_distance` can also be set. This means that if a cell has a cost that goes past this max then noData will be set as the cell's value in the resulting layer.\n",
```
jpolchlo: Still a bit unclear.  A suggested fix:

_`cost_distance` is an iterative method for approximating the weighted distance from a raster cell to a given geometry.  The `cost_distance` function takes in a geometry and a “friction layer” which essentially describes how difficult it is to traverse each raster cell.  Cells that fall within the geometry have a final cost of zero, while friction cells that contain noData values will correspond to noData values in the final result.  All other cells have a value that describes the minimum cost of traversing from  that cell to the geometry.  If the friction layer is uniform, this function approximates the Euclidean distance, modulo some scalar value._

It might still be useful to understand how noData values interact with this algorithm.  Do they become essentially uncrossable locations?  If we had a point on an island and noData values for the water cells, would we see noData values or defined values on the adjacent island?

docs/notebook-demos/map-algebra.ipynb
```diff
@@ -0,0 +1,655 @@
...
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "raster_poly_1 = box(0.0, 0.0, 5.0, 10.0)\n",
+    "raster_poly_2 = box(3.0, 6.0, 15.0, 20.0)\n",
+    "raster_poly_3 = box(13.5, 17.0, 30.0, 20.0)\n",
+    "\n",
+    "raster_mulit_poly = MultiPolygon([raster_poly_1, raster_poly_2, raster_poly_3])"
```
rossbernet: believe this should be `multi`

docs/notebook-demos/map-algebra.ipynb
jbouffard: That does read better, I'll replace what I had with that. I'm not really sure how the algorithm would behave in that instance, as I'm not familiar with how it's implemented in GeoTrellis.

docs/notebook-demos/map-algebra.ipynb
```diff
@@ -0,0 +1,655 @@
...
+    "tiled_layer = raster_layer.tile_to_layout(layout=gps.LocalLayout(tile_size=5))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Map Algebra\n",
+    "\n",
+    "Given a set of raster layers, it may be desirable to combine and filter the content of those layers.  This is the function of _map algebra_.  Two classes of map algebra operations are provided by GeoPySpark: _local_ and _focal_ operations.  Local operations individually consider the pixels or cells of one or more rasters, applying a function to the corresponding cell values.  For example, adding two rasters' pixel values to form a new layer is a local operation.\n",
```
jbouffard: I'm not sure if an introduction to map algebra is needed because of GeoPySpark's target audience should be somewhat familiar with the concept.

docs/notebook-demos/map-algebra.ipynb
```diff
@@ -0,0 +1,655 @@
...
+    "\n",
+    "Focal operations consider a region around each pixel of an input raster and apply an operation to each region.  The result of that operation is stored in the corresponding pixel of the output raster.  For example, one might weight a 5x5 region centered at a pixel according to a 2d Gaussian to effect a blurring of the input raster.  One might consider this roughly equivalent to a 2d convolution operation.\n",
```
jbouffard: Same as above, the gifs are good but I'm not sure what it would bring to someone whose already familiar with map algebra.

PR #477, opened 2017-08-15
Separate Testing for Each Test on Travis

jbouffard:
This PR adds a script that will run each unittest separately on Travis. This will _hopefully_  resolve the frequent build failures that occur due to memory issues on Travis.

### DIFF
--- a/test_script.sh
+++ b/test_script.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+testpath="geopyspark/tests/"
+testpattern="*_test.py"
+
+for f in $testpath$testpattern $testpath/**/$testpattern;
+do
+  pytest $f
+done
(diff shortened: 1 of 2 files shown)


test_script.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+testpath="geopyspark/tests/"
+testpattern="*_test.py"
+
+for f in $testpath$testpattern $testpath/**/$testpattern;
```
jamesmcclain: Do `$testpath$testpattern` and `$testpath/**/$testpattern` produce the same list or different ones?

Assuming that these two statement are not redundant, :+1:

jbouffard: They produce different ones. `$testpath$testpattern` retrieves all of the files in the root directory and `$testpaht/**/$testpattern` gets all of the files from the sub directories.

jamesmcclain: :+1:

PR #483, opened 2017-08-21
Visualization Notebook and Docs

jbouffard:
This PR will add the visualization notebooks and docs to GeoPySpark. 

**Note**: These notebooks and docs will not contain info on the `tms` module as I'm not familiar enough with it. Rather, `Pyramid` and `ColorMap` will be the focus of the notebook and docs.

### DIFF
--- a/docs/guides/visualization.rst
+++ b/docs/guides/visualization.rst
@@ -1,0 +1,279 @@
+.. _visualizing:
+
+Visualizing Data in GeoPySpark
+==============================
+
+Data is visualized in GeoPySpark by running a server which allows it to
+be viewed in an interactive way. Before putting the data on the server,
+however, it must first be formatted and colored. This guide seeks to go
+over the steps needed to create a visualization server in GeoPySpark.
+
+Before begining, all examples in this guide need the following boilerplate
+code:
+
+.. code:: python3
+
+    !curl -o /tmp/cropped.tif https://s3.amazonaws.com/geopyspark-test/example-files/cropped.tif
+
+.. code:: python3
+
+    import geopyspark as gps
+    import matplotlib.pyplot as plt
+
+    from colortools import Color
+    from pyspark import SparkContext
+
+    %matplotlib inline
+
+    conf = gps.geopyspark_conf(master="local[*]", appName="visualization")
+    pysc = SparkContext(conf=conf)
+
+    raster_layer = gps.geotiff.get(layer_type=gps.LayerType.SPATIAL, uri="/tmp/cropped.tif")
+    tiled_layer = raster_layer.tile_to_layout(layout=gps.GlobalLayout(), target_crs=3857)
+
+
+.. _pyramid:
+
+Pyramid
+-------
+
+The :class:`~geopyspark.Pyramid` class represents a list of ``TiledRasterLayer``\ s
+that represent the same area where each layer is a level within the pyramid
+at a specific zoom level. Thus, as one moves up the pyramid (starting a
+level 0), the image will have its pixel resolution increased by a power of 2
+for each level. It is this varying level of detail that allows an
+interactive tile server to be created from a ``Pyramid``. This class is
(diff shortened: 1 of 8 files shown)


jpolchlo:
Would this notebook on the TMS functions suffice? https://gist.github.com/jpolchlo/70798bac6a71235866841bda99c262cb

jbouffard:
Yeah, I think that if it was reformatted it would be good for the TMS section. Though, I think there needs to be more on how to use the TMS in a non GeoNotebook setting; as GeoPySpark can work independently of it.

docs/guides/visualization.rst
```diff
@@ -0,0 +1,287 @@
...
+
+.. _pyramid:
+
+Pyramid
+-------
+
+The ``pyramid`` class represents a list of ``TiledRasterLayer``\ s that
+represent the same area where each layer is a level within the pyramid
+at a specific zoom level. Thus, as one moves up the pyramid (starting a
+level 0), the image will become more zoomed by a power of 2 for each
```
echeipesh: `more zoomed` It would be better to be more precise here: "the pixel resolution will increase by power of two"

docs/guides/visualization.rst
```diff
@@ -0,0 +1,287 @@
...
+Through the ``TiledRasterLayer.pyramid`` method or by constructing it by
+passing in a ``[TiledRasterLayer]`` or
+``{zoom_level: TiledRasterLayer}`` to ``Pyramid``.
+
+Any ``TiledRasterLayer`` with a ``max_zoom`` can be pyramided. However,
+the resulting ``Pyramid`` may have limited functionality depending on
+the layout of the source ``TiledRasterLayer``. In order to be used for
+visualization, the ``Pyramid`` **must** have been created from
+``TiledRasterLayer`` that was tiled using a ``GlobalLayout`` and whose
+tile sizes are a power of 2.
```
echeipesh: Again in effort to be more precise, tile sizes do not change from level to level in the pyramid, only the spatial resolution of those tiles. That is any given pixel of a tile covers larger and larger area as you go up the pyramid, decreasing in the resolution of the tile.

docs/guides/visualization.rst
```diff
@@ -0,0 +1,287 @@
...
+
+.. code:: python3
+
+    gps.Pyramid({x: tiled_layer.tile_to_layout(gps.GlobalLayout(zoom=x)) for x in range(0, 13)})
+
+Computing the Histogram of a Pyramid
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+One can produce a ``Histogram`` instance representing all of the layers
+within a ``Pyramid`` via the ``get_histogram`` method.
```
echeipesh: I believe it takes the histogram of only the bottom most layer and uses that as representative histogram of the pyramid.

docs/guides/visualization.rst
```diff
@@ -0,0 +1,287 @@
...
+    hist = pyramided.get_histogram()
+    hist
+
+RDD Methods
+~~~~~~~~~~~
+
+``Pyramid`` contains methods for working with the ``RDD``\ s contained
+within its ``TiledRasterLayer``\ s. A list of these can be found
+here :ref:`rdd-methods`. When used, all internal ``RDD``\ s
+will be operated on.
```
echeipesh: This change would be out of scope for this PR but would like to point out that we're making a transition from talking about these things are `RDD`s, which they're deep down, to `Layer` which they are conceptually. This is what prompted renaming from `TiledRasterRDD` to `TiledRasterLayer` and in general we should avoid leaking the implementation details into docs unless we're specifically talking about them or their consequences (performance, error messages, execution environment)

jbouffard: That's a good point. I do agree that is outside of the scope of this PR, though. I'm going to create an issue that addresses this and other problems with the docs now.

PR #491, opened 2017-09-07
Add AttributeStore to catalog guide

echeipesh:


docs/guides/catalog.rst
```diff
@@ -402,3 +402,130 @@ Querying by Space and Time
...
+AttributeStore
+--------------
+
+When writing a layer GeoTrellis uses an ``AttributeStore`` to write layer metadata required to read and query the layer later.
+This class can be used outside of catalog ``write`` and ``query`` functions to inspect available layers and store additional, user defined, attributes.
+
+Creating AttributeStore
+~~~~~~~~~~~~~~~~~~~~~~~
+
+``AttributeStore`` can be created from the same ``URI`` that is given to ``write`` and ``query`` functions.
```
jbouffard: It might be good to have links to the `AttributeStore`'s [class docs](https://geopyspark.readthedocs.io/en/latest/docs/geopyspark.geotrellis.html#geopyspark.geotrellis.catalog.AttributeStore). We do this throughout the rest of the docs.

The following is how you'd create a link to that in the doc:
```
:class:`~geopyspark.geotrellis.catalog.AttributeStore`
```

docs/guides/catalog.rst
```diff
@@ -406,7 +406,7 @@ Querying by Space and Time
 AttributeStore
 --------------
 
-When writing a layer GeoTrellis uses an ``AttributeStore`` to write layer metadata required to read and query the layer later.
+When writing a layer, GeoTrellis uses an :class:`~geopyspark.geotrellis.catalog.AttributeStore` to write layer metadata required to read and query the layer later.
```
jbouffard: This should probably stay how it was since it's referring to GeoTrellis' `AttributeStore` and not GeoPySpark's

echeipesh: Thats a good point. On reflection its probably better to simplify and talk about GeoPySpark, since that still true and more consistent.

PR #494, opened 2017-09-12
LayerType Creation Isuse Fix

### LINKED ISSUE #492: TiledRasterLayer.layer_type is not `gps.LayerType`
```python
v = elev_tiled_rdd.lookup(342,787)
plt.imshow(v[0]['data'][0])
```
... this results in

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-70-a061cff6e096> in <module>()
----> 1 v = elev_tiled_rdd.lookup(342,787)
      2 plt.imshow(v[0]['data'][0])

/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/geotrellis/layer.py in lookup(self, col, row)
   1077 
   1078         if self.layer_type != LayerType.SPATIAL:
-> 1079             raise ValueError("Only TiledRasterLayers with a layer_type of Spatial can use lookup()")
   1080 
   1081         bounds = self.layer_metadata.bounds

ValueError: Only TiledRasterLayers with a layer_type of Spatial can use lookup()
```

jbouffard:
This PR resolves an issue that can occur during the creation of a layer where the resulting`layer_type` is not an instance of `LayerType`. This occurred in `geotiff.get` due to the following line:

https://github.com/locationtech-labs/geopyspark/blob/686ad28724d7648bec5e7e1a39cd48d5d80fca79/geopyspark/geotrellis/geotiff.py#L93

In addition to `geotiff.get`, this problem could also arise when creating the layer via the `from_numpy_rdd` methods, so those have been fixed as well in this PR.

This PR resolves #492

### DIFF
--- a/geopyspark/geotrellis/layer.py
+++ b/geopyspark/geotrellis/layer.py
@@ -315,7 +315,7 @@ class RasterLayer(CachableLayer, TileLayer):
     def __init__(self, layer_type, srdd):
         CachableLayer.__init__(self)
         self.pysc = get_spark_context()
-        self.layer_type = layer_type
+        self.layer_type = LayerType(layer_type)
         self.srdd = srdd
 
     @classmethod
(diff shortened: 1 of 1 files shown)


echeipesh:
It may be more fool proof to just apply this uniformly in the `__init__`: https://github.com/locationtech-labs/geopyspark/pull/494/files#diff-816135500070ccd4dc7eb21e66211aceL738

Given `LayerType(LayerType('spatial'))` should still be just `LayerType`

PR #500, opened 2017-09-19
Union Function

jbouffard:
This PR adds the `union` module and function to geopyspark. This will make it easier to union together layers as the user will no longer have to bring the `RDD`s over to python in order to perform the operation.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ProjectedRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ProjectedRasterLayer.scala
@@ -1,6 +1,7 @@
 package geopyspark.geotrellis
 
 import GeoTrellisUtils._
+
 import geotrellis.proj4._
 import geotrellis.raster._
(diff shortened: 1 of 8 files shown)


jpolchlo:
A question: are there well-defined expectations for when two overlapping TiledRasterLayers are unioned?

jbouffard:
@jpolchlo If those `TiledRasterLayer`s have the same keys, then the resulting `TiledRasterLayer` will contain duplicate keys - one from each of the original layers.

jpolchlo:
Do you think it's worth mentioning that behavior in the docs?  I don't know that the union operator defined by spark is necessarily obvious in its function to non-spark users, and you've basically just wrapped that functionality.

I also wonder if someone would want to know how to combine tile layers to get a single tile per key as well?  A much looser suggestion, though.

jbouffard:
@jpolchlo Yeah, I think that'd be worth mentioning. As of right now, `aggregate_by_cell` would be the only tool they'd have to remove the duplicates. However, we should probably offer more means of doing so in the future.

PR #503, opened 2017-09-20
RasterLayer.merge Method

### LINKED ISSUE #502: Create Merge Method for Layers
As of right now, there aren't many ways of dealing with duplicate keys in a layer. One could use `aggregate_by_cell` if one has a `TiledRasterLayer`, but otherwise the choices are limited. Not only is this an API issue, but it causes problems elsewhere.

For example, in the `toSpatialLayer()` method, all of the duplicate keys are merged together by default. However, this may not always be the desired outcome. Instead, the user should have a choice on whether the values should be merged or not.

https://github.com/locationtech-labs/geopyspark/blob/22976e5ba2dd5183bcf9ff5e160aa3eb207b9adb/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala#L380

jbouffard:
This PR adds the `merge_duplicate_keys` method to `RasterLayer` and `TiledRasterLayer`. This will provide an easier means to getting rid of duplicate keys within a layer. In addition to the new method, `to_spatial_layer` now has the `merge_duplicates` parameter which will determine if the duplicate keys should be merged once they converted to `SpatialKey`s.

This PR resolves #502

### DIFF
--- a/geopyspark/geotrellis/layer.py
+++ b/geopyspark/geotrellis/layer.py
@@ -128,14 +128,14 @@ def _reproject(target_crs, layout, resample_method, layer):
         raise TypeError("%s can not be used as target layout." % layout)
 
 
-def _to_spatial_layer(layer, target_time, merge_duplicates):
+def _to_spatial_layer(layer, target_time):
     if layer.layer_type == LayerType.SPATIAL:
         raise ValueError("The given already has a layer_type of LayerType.SPATIAL")
 
     if target_time:
-        return layer.srdd.toSpatialLayer(_convert_to_unix_time(target_time), merge_duplicates)
+        return layer.srdd.toSpatialLayer(_convert_to_unix_time(target_time))
     else:
-        return layer.srdd.toSpatialLayer(merge_duplicates)
+        return layer.srdd.toSpatialLayer()
 
 
 class TileLayer(object):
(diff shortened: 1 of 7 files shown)


geopyspark/geotrellis/layer.py
```diff
@@ -435,14 +435,18 @@ def to_geotiff_rdd(self,
 
         return create_python_rdd(result, ser)
 
-    def to_spatial_layer(self, target_time=None):
+    def to_spatial_layer(self, target_time=None, merge_duplicates=False):
```
echeipesh: I want to push back on this parameter here, it does not do anything different from subsequent `. merge_duplicate_keys` call and is a lot less clear because its a boolean.

It would result in lines like
```py
layer.to_spatial_layer(target_time, True)
```
vs 
```py
layer.to_spatial_layer(target_time). merge_duplicate_keys()
```

geopyspark/geotrellis/layer.py
```diff
@@ -577,6 +581,32 @@ def collect_keys(self):
         else:
             return [temporal_projected_extent_decoder(key) for key in self.srdd.collectKeys()]
 
+    def merge_duplicate_keys(self, num_partitions=None):
```
echeipesh: The name is a little loose and long, we're not merging keys, we're merging tiles with duplicate keys. We know they're duplicate because they have the same key. 

A little tough to come up with a name for it, so why not keep `merge`? Could there be a confusion that it means something else? (<- not rhetorical question, question question)

Other likely use for merge are two merge two layers together:

```py
layer1.merge() # merges duplicate keys

layer1 = catalog.query(...)
layer2 = catalog.query(...)
layer1.merge(layer2) # merges tiles from layer2 into tiles from layer1
```

PR #505, opened 2017-09-25
Point Value Lookups

### LINKED ISSUE #504: Sampling layer points
Given a layer we may want to sample point values. Doing lookups we may want to sample multiple points at once.

```python
layer = catalog.get("s3://....")

geojson = read_some_file("...")
point = shapley.read(geojson)
layer.point_value(point)
# [ (point, 3.2) ]
# [ (datetime1, 3.2), (datetime2, 3.1) ]

geojson = read_some_file("...")
points = shapley.read(geojson)
layer.point_value(points)
# if spatial: [ (point1, 3.2), (point2, 1) ]
# if temporal: [ (datetime1, point1, 3.2), (datetime2, point1, 3.1), (datetime1, point2, 5.2), (datetime2, point2, 5.1)]
```

The return type is a concern:
- It must be an array because we may have tiles with duplicate keys at any point.
- It would be good to encase them in a Feature, geometry plus some property.

jbouffard:
This PR adds the `get_point_values` method to `TiledRasterLayer`. This will allow for the lookup of values at a given point for `TiledRasterLayer`s. In addition to the new method, `SpatialPoint` and `SpaceTimePoint` are two new data structures that are used to represent the method's output.

This PR resolves #504

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
@@ -15,7 +15,7 @@ import geotrellis.raster.io.geotiff._
 import geotrellis.raster.io.geotiff.compression._
 import geotrellis.raster.rasterize._
 import geotrellis.raster.render._
-import geotrellis.raster.resample.ResampleMethod
+import geotrellis.raster.resample.{ResampleMethod, PointResampleMethod, Resample}
 import geotrellis.spark._
 import geotrellis.spark.costdistance.IterativeCostDistance
 import geotrellis.spark.io._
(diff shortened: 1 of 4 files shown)


geopyspark/geotrellis/__init__.py
```diff
@@ -394,6 +394,33 @@ def __new__(cls, tile_size=None, tile_cols=None, tile_rows=None):
 RasterizerOptions.__new__.__defaults__ = (True, 'PixelIsPoint')
 
 
+SpatialPoint = namedtuple("SpatialPoint", 'point values')
```
jbouffard: @echeipesh I want to get your opinion on the `SpatialPoint` and `SpaceTimePoint` types. Currently, they are simply wrappers around `(Point, [int or float])` pairs and don't serve much of a purpose beyond that. Unless we plan on expanding these types and their uses, I think it might be easier to just return a `{point: [int or float]}` instead. Because when I was working with these types, I found they'd be better suited as `dict`s in their current state.

PR #518, opened 2017-10-18
Filter By Times

jbouffard:
This PR adds the `filter_by_times` method to `RasterLayer` and `TiledRasterLayer`. This will allow any layer of type `SPACETIME` to filter their elements by times.

### DIFF
--- a/geopyspark/geotrellis/layer.py
+++ b/geopyspark/geotrellis/layer.py
@@ -6,6 +6,7 @@ when performing operations.
 import json
 import datetime
 import shapely.wkb
+import pytz
 from shapely.geometry import Polygon, MultiPolygon, Point
 from geopyspark.geotrellis.protobufcodecs import (multibandtile_decoder,
                                                   projected_extent_decoder,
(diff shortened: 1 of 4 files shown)


geopyspark/geotrellis/layer.py
```diff
@@ -724,6 +725,39 @@ def __str__(self):
...
+            time_intervals (``[datetime.datetime]``): A list of the time intervals to query.
+                This list can have one or multiple elements. If just a single element, then only
+                exact matches with that given time will be kept. If there are multiple times given,
+                then they are each paired together so that they form ranges of time. In the case
+                where there are an odd number of elements, then the remaining time will be treated
+                as a single query and not a range.
+
+        Note:
+            If nothing intersects the given ``time_intervals``, then the returned ``RasterLayer``
+            will be empty.
```
echeipesh: Just to clarify if I'm giving `time_intervals` as `[a, b, c, d]` then the query is going to be for intervals:
`[ (a,b), (b,c), (c,d)]` ?

jbouffard: The intervals would be `[ (a, b), (c, d) ]`. I used the `grouped(2)` method on the Scala `Array` to group them together.

PR #525, opened 2017-11-02
Fixed Bug in tileToLayout for TiledRasterLayers

### LINKED ISSUE #523: Rewrite the "Querying a Layer" Section of the Catalog Documentation so it is More Clear
Going over the `Querying a Layer` section of the catalog documentation, there are some incorrect comments and certain concepts are not clear. Mainly, what will be returned and why when querying with a geometry.

### LINKED ISSUE #524: Fix tileToLayout Logic so That It Uses the Correct MapKeyTransform
We have recently found instances where `tile_to_layout` does not produce the correct number of elements in the resulting layer. This is because when we transform the keys of the layer into extents, we do so using the `MapKeyTransform` of the target the layout, and not the source. To resolve this, all we'd need to do is use the `MapKeyTransform` from the layout of the source. Which would look something like this:

```
val baseMapTransform = rdd.metadata.layout.mapTransform

rdd.map { case(k, v) => (ProjectedExtent(baseMapTransform(k), crs), v) }
```

jbouffard:
This PR fixes a bug present in the `tileToLayout` methods for the `TiledRasterLayer` classes by using the correct `MapKeyTransform` to convert the keys to extents.

This PR resolves #524

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
@@ -97,17 +97,15 @@ class SpatialTiledRasterLayer(
     zoom: Option[Int],
     resampleMethod: ResampleMethod
   ): TiledRasterLayer[SpatialKey] = {
-    val mapKeyTransform =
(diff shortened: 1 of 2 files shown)


jbouffard:
@echeipesh Okay, I'll make an issue to have that method moved to the abstract class.

PR #533, opened 2017-11-07
Layer Serialization

jbouffard:
This PR makes `RasterLayer` and `TiledRasterLayer` serializable. This should make all layer classes serializable


echeipesh:
What prompted this? I'm concerned that the spark closures might be grabbing more than they need if they're reaching for these classes.

If it is justified it would probably be good to mark the `rdd` val in the subclasses as `transient` to prevent it from being pulled into serialization.

jbouffard:
This was done because I was having serialization issues when going to and from Python `RDD`s on EMR. To resolve that issue, I initially just had all of the `Layer` classes extend `Serializable`, but thought just having `RasterLayer` and `TiledRasterLayer` extend `Serializable` would make things a little more centralized.

PR #541, opened 2017-11-10
combine_bands Method for RasterLayers and TiledRasterLayers

### LINKED ISSUE #539: Expose the combineValues method to GeoPySpark
We should add a modified [combineValues](https://github.com/locationtech/geotrellis/blob/2419b48e7f3ab848c4dafe8cbd3e6fb2522ff298/spark/src/main/scala/geotrellis/spark/mapalgebra/CombineMethods.scala#L27) method to the layer classes. While it won't be as robust as the GeoTrellis version, we could still use it to do many useful things, For example, combining bands of one and `n` RDDs so that they produce an new value which has `n + 1` bands. This would be useful when working with datasets where each band of an image is its own file (ie. Landsat).

jbouffard:
This PR will allow the user to combine bands from different layers so that the resulting layer will have a `bandCount` equal to the sum of the `bandCount`s of the input layers. For example, if `layer a` has 2 bands and `layer b` has one, then the resulting layer will have a `bandCount` of 3. Where 2 of the bands come from `layer a` and 1 comes from `layer b`.

This PR resolves #539

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
@@ -13,6 +13,7 @@ import geotrellis.spark._
 import geotrellis.spark.io._
 import geotrellis.spark.tiling._
 import geotrellis.vector._
+import org.apache.spark._
 import org.apache.spark.api.java.JavaRDD
 import org.apache.spark.rdd._
 import org.apache.spark.storage.StorageLevel
@@ -26,6 +27,8 @@ import spire.std.any._
 import scala.reflect.{ClassTag, classTag}
 import scala.collection.JavaConverters._
 import scala.util.Try
+
+import java.util.ArrayList
 
 abstract class TileLayer[K: ClassTag] {
   def rdd: RDD[(K, MultibandTile)]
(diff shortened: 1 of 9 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
```diff
@@ -239,4 +242,42 @@ object TileLayer {
...
+          bandMap + (value._1 -> value._2.bands),
+        (m1: Map[Int, Vector[Tile]], m2: Map[Int, Vector[Tile]]) =>
+          m1 ++ m2
+      )
+
+    bands.mapValues { case (v: Map[Int, Vector[Tile]]) =>
+      MultibandTile(
+        v.toSeq
+          .sortWith(_._1 < _._1)
+          .toMap
```
echeipesh: I believe turning the collection back to map may undo the sort and isn't actually required step here.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
```diff
@@ -239,4 +242,42 @@ object TileLayer {
...
+    val arr = Array.ofDim[RDD[(K, (Int, MultibandTile))]](rdds.size)
+
+    for ((layer, index) <- rdds.zipWithIndex) {
+      arr(index) = layer.mapValues { (index, _) }
+    }
+
+    val unioned = sc.union(arr.toSeq)
+
+    val bands: RDD[(K, Map[Int, Vector[Tile]])] =
+      unioned.combineByKey(
```
echeipesh: There should be some kind of check for error or warning if the layouts of the different layers do not match. Thats a kind of error thats easy to do with a surprising result.

geopyspark/geotrellis/union.py
```diff
@@ -26,37 +26,16 @@ def union(*layers):
     """
 
     if len(layers) == 1:
-        raise ValueError("Union can only performed on 2 or more layers")
+        raise ValueError("combine_bands can only performed on 2 or more layers")
```
echeipesh: Search/replace error ?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
jbouffard: I had the layouts checked in python: https://github.com/jbouffard/geopyspark/blob/acb4163cbbcfdd8ad6dc3adceef5a9c6bf96effc/geopyspark/geotrellis/__init__.py#L66

geopyspark/geotrellis/union.py
jbouffard: Yeah, that's definitely a search/replace error.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
jbouffard: Ah, I didn't know that. I'll fix it.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TileLayer.scala
echeipesh: Gotcha

PR #542, opened 2017-11-10
Better Temporal Resolution Control When Writing Layer

### LINKED ISSUE #538: Further Specifiy the Resolution of a Temporal Layer
When saving a temporal layer, there is no way of specifying how many of a particular time unit should be grouped together. For example, if using a temporal resolution of days, then it'll save each day as its own file. However, if a user wanted to group the days so that there was more than day per file, then there'd be no way of doing that.

The exact API to handle this may be a bit tricky, so this issue tackled carefully.

jbouffard:
This PR adds the `temporal_resolution` parameter to the `write` method. This will allow users to save their layers with a more fine-grained resolution than what was previously available.

This PR resolves #538

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerWriterWrapper.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerWriterWrapper.scala
@@ -43,18 +43,27 @@ class LayerWriterWrapper(attributeStore: AttributeStore, uri: String) {
 
   private def getTemporalIndexMethod(
     timeString: String,
+    timeResolution: String,
     indexStrategy: String
   ) =
-    (indexStrategy, timeString) match {
-      case ("zorder", "millis") => ZCurveKeyIndexMethod.byMilliseconds(1)
-      case ("zorder", "seconds") => ZCurveKeyIndexMethod.bySecond
-      case ("zorder", "minutes") => ZCurveKeyIndexMethod.byMinute
-      case ("zorder", "hours") => ZCurveKeyIndexMethod.byHour
-      case ("zorder", "days") => ZCurveKeyIndexMethod.byDay
-      case ("zorder", "weeks") => ZCurveKeyIndexMethod.byDays(7)
-      case ("zorder", "months") => ZCurveKeyIndexMethod.byMonth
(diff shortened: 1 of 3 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerWriterWrapper.scala
```diff
@@ -43,17 +43,27 @@ class LayerWriterWrapper(attributeStore: AttributeStore, uri: String) {
 
   private def getTemporalIndexMethod(
     timeString: String,
+    timeResolution: String,
```
echeipesh: Why `String`? Seems like this would always have to be a number, an int/long really.

geopyspark/geotrellis/catalog.py
```diff
@@ -250,6 +251,15 @@ def write(uri,
             each index. Meaning, what time intervals are used to seperate each record. While this is
             set to ``None`` as default, it must be set if saving spatial-temporal data.
             Depending on the indexing method chosen, different time units are used.
+        time_resolution (str or int, optional): Determines how data for each ``time_unit`` should be
```
echeipesh: `time_resolution` sounds like a the in-house name for this, struggling to come up with a more intuitive alternative though. Fishing for ideas here.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/io/LayerWriterWrapper.scala
jbouffard: I did `String` just because it seemed easier than having to deal with both `Int` and `Long` as a possible input

geopyspark/geotrellis/catalog.py
jbouffard: What about `temporal_grouping`? Since it's used to determine how the units of time should be grouped together when saved.

echeipesh: Maybe `time_window` or `time_span` if those don't inspire clarity in you we can leave it as is.

jbouffard: The only issue I have with those is that they could be interpreted to mean something different. Like, they control the the time frame of the data being saved (ie. only data falling within this window/span will be saved).

PR #552, opened 2017-11-21
EmptyQueryException For LayerReaderWrapper

### LINKED ISSUE #531: Return a More Helpful Error Message When Querried Geometry Does Not Intersect Layer
Right now, when querying a layer with a geometry the user will get an unhelpful error message if no intersection occurs. This error is an `empty.reduceLeft` error, and is unhelpful. There should be an additional check step before the `reduce` to ensure that the collection is not empty

jbouffard:
This PR will now throw an `EmptyQueryException` when the `layerQuery` does not intersect the chosen layer. Before, the error that was thrown was a `java.lang.UnsupportedOperationException: empty.reduceLeft` which was unhelpful.

This PR resolves #531

### DIFF
--- a/docs/guides/catalog.rst
+++ b/docs/guides/catalog.rst
@@ -303,7 +303,7 @@ an :class:`~geopyspark.geotrellis.Extent`.
 
 **Note**: It is important that the given geometry is in the same
 projection as the queried layer. Otherwise, either the wrong area
-will be returned or an exception will be thrown.
+will be returned or an empty layer will be returned.
 
 When the Queried Geometry is in the Same Projection as the Layer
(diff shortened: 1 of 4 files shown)


echeipesh:
Exception on empty query doesn't seem helpful. IIRC GT will return an empty RDD for non-intersecting query. I suspect the error on `reduceLeft` was from GPS backend?

Overall I would expect to receive either equivalent of empty layer or `None`.

jbouffard:
@echeipesh Yeah, the error was caused by the `reduceLeft` on an empty collection. It will now return an empty layer if the queried area doesn't intersect. I always added a bit more resources to make it easier to determine if the layer is empty.

Issue #466, opened 2017-08-12
Error reading Landsat Data off S3


lossyrob:
This is either a Bug or a confusing error message:

<img width="1131" alt="screen shot 2017-08-12 at 7 37 01 pm" src="https://user-images.githubusercontent.com/2320142/29244935-dac7b19e-7f95-11e7-9c5e-a49720286606.png">

Code:
```python
import geopyspark as gps

from pyspark import SparkContext
# Set up our spark context
conf = gps.geopyspark_conf(appName="Landsat") \
          .setMaster("local[*]") \
          .set(key='spark.ui.enabled', value='true') \
          .set(key="spark.driver.memory", value="8G")
sc = SparkContext(conf=conf)
target_image = "s3://landsat-pds/c1/L8/138/217/LC08_L1GT_138217_20170703_20170703_01_RT/LC08_L1GT_138217_20170703_20170703_01_RT_B5.TIF"
rdd = gps.geotiff.get(gps.LayerType.SPATIAL, target_image, max_tile_size=512, num_partitions=50)
tiled_layer = rdd.tile_to_layout(gps.GlobalLayout(), target_crs="EPSG:3857")
```

Full stack trace:
```
---------------------------------------------------------------------------
Py4JJavaError                             Traceback (most recent call last)
<ipython-input-12-9d091f0366da> in <module>()
----> 1 tiled_layer = rdd.tile_to_layout(gps.GlobalLayout(), target_crs="EPSG:3857")

/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/geotrellis/layer.py in tile_to_layout(self, layout, target_crs, resample_method)
    644         if target_crs:
    645             target_crs = crs_to_proj4(target_crs)
--> 646             return _reproject(target_crs, layout, resample_method, self)
    647 
    648         if isinstance(layout, Metadata):

/home/hadoop/.local/lib/python3.4/site-packages/geopyspark/geotrellis/layer.py in _reproject(target_crs, layout, resample_method, layer)
    108 
    109     if isinstance(layout, (LocalLayout, GlobalLayout, LayoutDefinition)):
--> 110         srdd = layer.srdd.reproject(target_crs, layout, resample_method)
    111         return TiledRasterLayer(layer.layer_type, srdd)
    112 

/usr/local/spark-2.1.0-bin-hadoop2.7/python/lib/py4j-0.10.4-src.zip/py4j/java_gateway.py in __call__(self, *args)
[...]

PR #561, opened 2017-11-28
aws-java-sdk-s3 Version Change

### LINKED ISSUE #485: Find True Cause of Failure When Reading GeoTiffs From S3
In #466, a problem was found where setting both `max_tile_size` and `num_partitions` in the `geotiff.get` function causes it to fail when it reads from S3. After a bit investigating, the source of this issue was found to be a change in version of aws' java sdk for S3. #482 was able to provide a temporary solution by overriding the S3 dependency to version 1.11.105. However, this doesn't resolve the actual issue, which is still unknown.

A breaking change occurred in version [1.11.106](https://github.com/aws/aws-sdk-java/blob/master/CHANGELOG.md#111106-2017-03-20), so these changes need to be looked into to determine the true cause of this bug.

jbouffard:
This PR changes the version of GeoPySpark's `aws-java-sdk-s3` dependency from 1.11.105 to the version used in GeoTrellis. This was originally changed because of issues encountered when reading from S3 #466. At the time, we were unsure of the issue and had decided to come back to it later.

While the exact cause is still not known, I can speculate that the logic in the `S3GeoTiffRDD` was the source. As this was the only thing that changed between now and when the issue first appeared.

This code below could not work before, but now can even with the updated `aws-java-sdk-s3` version:

```
In [1]: import geopyspark as gps
   ...: 
   ...: from pyspark import SparkContext
   ...: # Set up our spark context
   ...: conf = gps.geopyspark_conf(appName="Landsat") \
   ...:           .setMaster("local[*]") \
   ...:           .set(key='spark.ui.enabled', value='true') \
   ...:           .set(key="spark.driver.memory", value="8G")
   ...: sc = SparkContext(conf=conf)
   ...: target_image = "s3://landsat-pds/c1/L8/138/217/LC08_L1GT_138217_20170703_20170703_01_RT/LC08_L1GT_138217_201
   ...: 70703_20170703_01_RT_B5.TIF"
   ...: rdd = gps.geotiff.get(gps.LayerType.SPATIAL, target_image, max_tile_size=512, num_partitions=50)
   ...: tiled_layer = rdd.tile_to_layout(gps.GlobalLayout(), target_crs="EPSG:3857")
   ...: 
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
17/11/28 13:47:44 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2017-11-28 13:47:45,613 WARN  - numPartitions option is ignored
17/11/28 13:47:45 WARN S3GeoTiffRDD$: numPartitions option is ignored
                                                                                
In [2]: tiled_layer
Out[2]: TiledRasterLayer(layer_type=LayerType.SPATIAL, zoom_level=13, is_floating_point_layer=False)
```

This PR resolves #485

### DIFF
--- a/geopyspark-backend/geotrellis/build.sbt
+++ b/geopyspark-backend/geotrellis/build.sbt
@@ -21,8 +21,6 @@ libraryDependencies ++= Seq(
   "net.sf.py4j"           % "py4j"                         % "0.10.5"
 )
 
-dependencyOverrides += "com.amazonaws" % "aws-java-sdk-s3" % "1.11.105"
-
 assemblyMergeStrategy in assembly := {
   case "reference.conf" => MergeStrategy.concat
   case "application.conf" => MergeStrategy.concat


echeipesh:
Weird that this has just gone away, suspect it has to do with refactors in `S3GeoTiffRDD` losing a troublesome usage.

jbouffard:
I suspect that as well. Though, I'm glad that the issue is resolved.

PR #566, opened 2017-12-01
Rasterize an RDD of Geometries

### LINKED ISSUE #521: Rasterize RDD of Geometries
Current implementation of `rasterize` method is takes a collection of Geometries and broadcasts them from the driver to create the RDD to be rasterized. There needs to be a version of this method that accepts a python RDD of geometries instead of creating one.

We have no notion of what an RDD of geometries looks like in pyspark. We should consider the representations provided by [GeoMesa pyspark bindings](https://github.com/locationtech/geomesa/blob/master/geomesa-spark/geomesa_pyspark/src/main/python/geomesa_pyspark/spark.py) and support at least one of them as common representation in the API.

jbouffard:
This PR will allow the user to pass in an `RDD` of `shapely.geometry`s to the `rasterize` method. In addition, it updates the documentation on this method to reflect these changes.

This PR resolves #521

### DIFF
--- a/geopyspark/geotrellis/rasterize.py
+++ b/geopyspark/geotrellis/rasterize.py
@@ -1,4 +1,4 @@
-import shapely.wkb
+from shapely.wkb import dumps
 from geopyspark import get_spark_context
 from geopyspark.geotrellis.constants import LayerType, CellType
 from geopyspark.geotrellis.layer import TiledRasterLayer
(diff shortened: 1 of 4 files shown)


geopyspark/geotrellis/rasterize.py
```diff
@@ -29,13 +30,33 @@ def rasterize(geoms, crs, zoom, fill_value, cell_type=CellType.FLOAT64, options=
...
+                          CellType(cell_type).value,
+                          options,
+                          num_partitions)
+
+    else:
+        wkb_rdd = geoms.map(lambda geom: dumps(geom))
+
+        # If this is False then the WKBs will be serialized
+        # when going to Scala resulting in garbage
+        wkb_rdd._bypass_serializer = True
```
echeipesh: Nice find. Looks like this bypasses the normal use of python serializers and allows the `JavaRDD` to get at the WKB bytes directly?

jbouffard: Thanks. Yeah, or else the WKBs will be serialized with what ever serializer that `RDD` has which is unhelpful in this case.

jbouffard:
@echeipesh That makes sense. Luckily, that shouldn't be too hard to implement. We could probably describe a `Feature` as:

```python
namedtuple("Feature", "geometry value")
```

PR #567, opened 2017-12-04
tobler Method for TiledRasterLayer

jbouffard:
This PR adds the `tobler` method to `TiledRasterLayer`. The logic used create the tobler was taken from this PR: https://github.com/locationtech/geotrellis/pull/2432

**Note**: As mentioned in the linked PR, the `tobler` method assumes all sloped are up hill, and thus, this can result incorrect results. This will be fixed in a future update.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
@@ -11,6 +11,7 @@ import geotrellis.raster._
 import geotrellis.raster.distance._
 import geotrellis.raster.io.geotiff._
 import geotrellis.raster.io.geotiff.compression._
+import geotrellis.raster.mapalgebra.focal.Square
 import geotrellis.raster.mapalgebra.local._
 import geotrellis.raster.rasterize._
 import geotrellis.raster.render._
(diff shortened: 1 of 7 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
```diff
@@ -314,6 +315,26 @@ abstract class TiledRasterLayer[K: SpatialComponent: JsonFormat: ClassTag: Bound
...
+  ): TiledRasterLayer[K] = {
+    val n = Square(1)
+    val target = TileLayer.getTarget(targetCell)
+    val singlebandRDD: TileLayerRDD[K] = rdd.withContext(_.mapValues(_.band(0)))
+    val cellSize = singlebandRDD.metadata.layout.cellSize
+
+    val toblerCalc = (tile: Tile, bounds: Option[GridBounds], cellSize: CellSize) =>
+      Tobler(tile, n, bounds, cellSize, zFactor, target)
+
+    val calc = (tile: Tile, bounds: Option[GridBounds]) => toblerCalc(tile, bounds, cellSize)
```
echeipesh: The indirection here seems strange. Why not call `Tobler` directly?

jbouffard: I didn't call `Tobler` directly because because I was using `FocalOperation` instead. If that's not clear, though, I could use it directly.

PR #568, opened 2017-12-04
Expose Partitioners in the API

jbouffard:
This PR exposes a choice of a `Partitioner` when performing operations in GPS. The two that have been added are: `HashPartitioner` and `SpatialPartitioner`, which was created by @echeipesh. As of this PR, the methods where the `Partitioner`s are exposed in are: `repartition`, `merge`, `pyramid`, and `rasterize`.

**Note**: The way to add custom `Partitioner`s in PySpark is by passing in `partitionFunc` as a parameter instead of a `Partition` instance (see [this](https://github.com/apache/spark/blob/7d039e0c0af0931a1696d89e76f455ae4adf277d/python/pyspark/rdd.py#L1736) as an example). Because of this, there might not be a way in which we can preserve the same partitioning strategy a user set when working with the layer in Scala.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/Constants.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/Constants.scala
@@ -60,4 +60,7 @@ object Constants {
 
   final val METERSATEQUATOR = 11320
   final val FEETATEQUATOR = 365217.6
+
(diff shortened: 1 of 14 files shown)


jbouffard:
@jamesmcclain We never have supported Python partitioners. The main reason being that PySpark doesn't have them. Instead, the user passes in a partitioning strategy function to certain methods. I'm not really sure how we could support those at the moment, though.

PR #584, opened 2017-12-11
Vector Pipe Integration Cleanup

jbouffard:
This PR cleans up the vector pipe code further by reorganizing the tests and removing unneeded files. In additions to the cleanup, the behavior for getting the tags from a `FeatureCollection` has changed so that an error won't be thrown if the desired collection is empty.


.travis.yml
```diff
@@ -30,6 +31,8 @@ install:
   - pushd geopyspark-backend &&
     ./sbt "project geotrellis-backend" assembly &&
     cp geotrellis/target/scala-2.11/geotrellis-backend-assembly-0.3.0.jar ../geopyspark/jars &&
+    ./sbt "project vectorpipe" assembly &&
+    cp vectorpipe/target/scala-2.11/vectorpipe-assembly-0.3.0.jar ../geopyspark/jars &&
```
echeipesh: Could the version be either and ENV var or wildcard so this doesn't need to be updated for every pump?

jbouffard: Yeah, I could implement something like that to make things easier.

PR #586, opened 2017-12-12
Fix Tobler and Slope

echeipesh:


jbouffard:
@echeipesh Could you please rebase this on top of the `feature/vector-pipe` branch?

jbouffard:
Sorry, I misunderstood what was going on. Could change the target branch to `feature:vector-pipe`?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
```diff
@@ -151,7 +151,12 @@ class SpatialTiledRasterLayer(
       rdd.metadata
     )
 
-    val _neighborhood = getNeighborhood(operation, neighborhood, param1, param2, param3)
+    val _neighborhood =
+      if (operation == Constants.SLOPE || operation == Constants.ASPECT)
+        getNeighborhood(operation, neighborhood, 1.0, 0.0, 0.0)
```
jbouffard: Are we now going to overwrite the size/area of a `Neighborhood` when the user wants to use a `SLOPE` or `ASPECT` operation? If so, then we'll have to update the docs to reflect that.

jbouffard: I know you didn't create the method, but looking at the [getNeighborhood](https://github.com/locationtech-labs/geopyspark/blob/fe5588d06ee3137705102a6f2af3ff6412b4280b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/GeoTrellisUtils.scala#L50) method, it looks like that `operation` parameter is not used at all and can be removed.

echeipesh: Hmm, actually `Slope` and `Aspect` are only defined for Square(1) and their implementations in GeoTrellis assume it, which makes them different from other focal operations.

That seems to imply that they should be treated different, perhaps with `.slope` and `.aspect` methods ?

jbouffard: I see. In that case we should probably break them out into their own methods. That doesn't need to happen for this PR, but I think we should break them out before merging the `feature/vector-pipe` branch into `master`.

PR #589, opened 2017-12-13
TMS Display Fix

jbouffard:
This PR fixes the issues with the TMS display where the same element would change colors (and therefore, seemingly values) as one zooms in and out of the map. Not setting the `ContentType` of the response appears to have been the cause of these color issues; so by specifying it, the problem seems resolved.


jpolchlo:
My Akka-http knowledge is failing me here, but I have a concern about your solution being a bit of a heavy hammer.  Might make it difficult to extend to different image formats, if such a thing were ever to become desirable.  I found https://stackoverflow.com/a/42118893/8468924, which describes how one might inspect the "accepts" header to negotiate the proper response type.  We can provide a content type using the method described at the top of https://stackoverflow.com/a/32201336/8468924.

You should for a response from @echeipesh to see if my concerns are overblown?  Maybe the marshaller here is required either way?

jbouffard:
@jpolchlo My original solution was to use the method described in your second link. However, I noticed that sometimes the server would stop working when `complete` was given an `HttpResponse`. It would throw a bunch of errors (which I unfortunately didn't save), and then it would completely shut down. Even if I `unbind`ed it and tried to `bind` another source, it still wouldn't work. I had to restart iPython in order to get the TMS server to run again.

My knowledge of Akka Http is very limited, though. So there's a chance I could have not been doing something correctly.

PR #595, opened 2017-12-15
Slope Method for TiledRasterLayer Classes

jbouffard:
This PR breaks out the `Slope` focal operation so that it is now its own method called, `slope`. In addition to breaking out `Slope`, the `ZFactorCalculator` class was created. This class is used to derive a `ZFactor` from a `Tile` and is needed for the `Slope` operation.

This PR cannot be merged until #584 is merged.


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/ZFactorCalculator.scala
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ZFactorCalculator(zFactorProducer: Double => Double) extends Serializable {
+  def deriveZFactor(extent: Extent): Double =
+    deriveZFactor(extent.center)
+
+  def deriveZFactor(center: Point): Double =
+    deriveZFactor(center.y)
+
+  def deriveZFactor(lat: Double): Double =
```
echeipesh: This should be `y` if its intended to be used with other projections.

geopyspark/geotrellis/constants.py
```diff
@@ -286,10 +285,8 @@ class Compression(Enum):
...
+class Unit(Enum):
+    """Represents the units of elevation."""
 
-class TargetCell(Enum):
-    """Determines which cells in the layer should be used when running a computation."""
-
-    ALL = "All"
-    DATA = "Data"
-    NODATA = "NoData"
+    METER = "Meters"
```
echeipesh: Should be `Meter` to keep with in sync with constant name

PR #602, opened 2018-01-04
local_max Opertion for TiledRasterLayers

### LINKED ISSUE #601: Implement a Local Max Method For TiledRasterLayer
Right now, there is no way to perform a local max operation on a `TiledRasterLayer`. We should add a `local_max` method that can take an `int`, `float`, or another `TiledRasterLayer` and produced a new layer with the resulting max cell values.

jbouffard:
This PR adds the `local_max` method to `TiledRasterLayer`. This will allow the user to perform local max operations using `int`s, `float`, or other `TiledRasterLayer`s as the max constant.

This PR resolves #601

### DIFF
--- a/geopyspark/geotrellis/layer.py
+++ b/geopyspark/geotrellis/layer.py
@@ -1842,6 +1842,28 @@ class TiledRasterLayer(CachableLayer, TileLayer):
 
         return TiledRasterLayer(self.layer_type, srdd)
 
+    def local_max(self, value):
+        """Determines the maximum value for each cell of each ``Tile`` in the layer.
+
+        This method takes a ``max_constant`` that is compared to each cell in the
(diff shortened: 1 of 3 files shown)


geopyspark/geotrellis/layer.py
```diff
@@ -1842,6 +1842,28 @@ def _process_operation(self, value, operation):
 
         return TiledRasterLayer(self.layer_type, srdd)
 
+    def local_max(self, max_constant):
```
echeipesh: `max_constant` is confusing name for something that can be a layer. Since it can be a constant or a layer I think you're down to something generic like `value` or `other`.

jbouffard: :+1:

PR #626, opened 2018-01-30
Ensure multiple concurrent notebooks can create TMS servers

jpolchlo:
When working with multiple notebooks that wish to use the TMS server, we were jumping the gun and starting the Py4J callback server using the wrong method and too soon.  This should fix the problem and avoid port collisions in the future.

I've tested that this change preserves functionality with the following script:
https://gist.github.com/jpolchlo/c9c432501fa47a276a6d719095753ee4#file-srtm-orc-tobler-display-tms-callback-test-ipynb


jbouffard:
This looks good to me :+1:

PR #629, opened 2018-02-02
Mask Layers By RDD[Geometry]

jbouffard:
This PR makes it so that users can mask a layer via an `RDD[shapely.geometry]`. There may be issues with the performance of this new feature, mainly when dealing with complex geometries. We should run benchmarks in the near future to make the needed changes if performance becomes an issue.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
@@ -210,6 +210,30 @@ class SpatialTiledRasterLayer(
   def mask(geometries: Seq[MultiPolygon]): TiledRasterLayer[SpatialKey] =
     SpatialTiledRasterLayer(zoomLevel, Mask(rdd, geometries, Mask.Options.DEFAULT))
 
+  def mask(
+    groupedRDD: RDD[(SpatialKey, Iterable[Geometry])],
+    options: Rasterizer.Options
+  ): TiledRasterLayer[SpatialKey] = {
+    val mapTrans = rdd.metadata.layout.mapTransform
+
+    val joinedRDD: RDD[(SpatialKey, (MultibandTile, Iterable[Geometry]))] =
+      groupedRDD.partitioner match {
+        case Some(p) =>
+          rdd.join(groupedRDD, p)
+        case None =>
+          rdd.join(groupedRDD)
+      }
+
+    val maskedRDD: RDD[(SpatialKey, MultibandTile)] =
+      joinedRDD.mapPartitions ({ partition =>
+        partition.map { case (k, (v, geoms)) =>
+          (k, v.mask(mapTrans(k), geoms, options))
+        }
+      }, preservesPartitioning = true)
+
+    SpatialTiledRasterLayer(zoomLevel, ContextRDD(maskedRDD, rdd.metadata))
+  }
+
   def stitch: Array[Byte] =
     PythonTranslator.toPython[MultibandTile, ProtoMultibandTile](ContextRDD(rdd, rdd.metadata).stitch.tile)
 
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
@@ -87,6 +87,33 @@ class TemporalTiledRasterLayer(
   def mask(geometries: Seq[MultiPolygon]): TiledRasterLayer[SpaceTimeKey] =
     TemporalTiledRasterLayer(zoomLevel, Mask(rdd, geometries, Mask.Options.DEFAULT))
 
+  def mask(
+    groupedRDD: RDD[(SpatialKey, Iterable[Geometry])],
+    options: Rasterizer.Options
+  ): TiledRasterLayer[SpaceTimeKey] = {
+    val mapTrans = rdd.metadata.layout.mapTransform
+
+    val rekeyedRDD: RDD[(SpatialKey, (SpaceTimeKey, MultibandTile))] =
+      rdd.map { case (k, v) => (k.getComponent[SpatialKey], (k, v)) }
+
+    val joinedRDD: RDD[(SpatialKey, ((SpaceTimeKey, MultibandTile), Iterable[Geometry]))] =
+      groupedRDD.partitioner match {
+        case Some(p) =>
+          rekeyedRDD.join(groupedRDD, p)
+        case None =>
+          rekeyedRDD.join(groupedRDD)
+      }
+
+    val maskedRDD: RDD[(SpaceTimeKey, MultibandTile)] =
+      joinedRDD.mapPartitions ({ partition =>
+        partition.map { case (k, ((tk, v), geoms)) =>
+          (tk, v.mask(mapTrans(k), geoms, options))
+        }
+      }, preservesPartitioning = true)
+
+    TemporalTiledRasterLayer(zoomLevel, ContextRDD(maskedRDD, rdd.metadata))
+  }
+
   private def wkbsToMultiPolygons(wkbs: java.util.ArrayList[Array[Byte]]) = {
     wkbs
       .asScala.map({ wkb => WKB.read(wkb) })
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
@@ -93,6 +93,35 @@ abstract class TiledRasterLayer[K: SpatialComponent: JsonFormat: ClassTag: Bound
     mask(geometries)
   }
 
+  def mask(
+    wkbRDD: RDD[Array[Byte]],
+    partitionStrategy: PartitionStrategy,
+    options: Rasterizer.Options
+  ): TiledRasterLayer[K] = {
+    val geometryRDD: RDD[Geometry] = wkbRDD.map { WKB.read }
+    val clipped: RDD[(SpatialKey, Geometry)] = geometryRDD.clipToGrid(rdd.metadata.layout)
+
+    val partitioner: Option[Partitioner] =
+      partitionStrategy match {
+        case ps: PartitionStrategy => ps.producePartitioner(rdd.getNumPartitions)
+        case null => None
+      }
+
+    val groupedRDD: RDD[(SpatialKey, Iterable[Geometry])] =
+      (partitioner, rdd.partitioner) match {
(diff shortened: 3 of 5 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
```diff
@@ -210,6 +210,31 @@ class SpatialTiledRasterLayer(
   def mask(geometries: Seq[MultiPolygon]): TiledRasterLayer[SpatialKey] =
     SpatialTiledRasterLayer(zoomLevel, Mask(rdd, geometries, Mask.Options.DEFAULT))
 
+  def mask(
+    aggregatedRDD: RDD[(SpatialKey, Seq[Geometry])],
+    producePartitioner: () => Option[Partitioner],
```
echeipesh: This seems to break the pattern, why is it a function rather than strategy ?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
```diff
@@ -87,6 +87,36 @@ class TemporalTiledRasterLayer(
...
+        case None =>
+          rekeyedRDD.join(aggregatedRDD)
+      }
+
+    val maskedRDD: RDD[(SpaceTimeKey, MultibandTile)] =
+      joinedRDD.mapPartitions ({ partition =>
+        partition.map { case (k, ((tk, v), geoms)) =>
+          (tk, v.mask(mapTrans(k), geoms, options))
+        }
+        // TODO: Make sure that preservesPartitioning doesn't negativly impact
```
echeipesh: What is the potential concern with it ?

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
```diff
@@ -210,6 +210,31 @@ class SpatialTiledRasterLayer(
...
+    val joinedRDD: RDD[(SpatialKey, (MultibandTile, Seq[Geometry]))] =
+      producePartitioner() match {
+        case Some(p) =>
+          rdd.join(aggregatedRDD, p)
+        case None =>
+          rdd.join(aggregatedRDD)
+      }
+
+    val maskedRDD: RDD[(SpatialKey, MultibandTile)] =
+      joinedRDD.mapPartitions ({ partition =>
```
echeipesh: We really should just make `.mapValuesWithKey` that implements this `.mapPartitions` `preservePartitioning` pattern as an implicit extension like `PairRDDFunctions`, we just do it so often.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
```diff
@@ -93,6 +93,45 @@ abstract class TiledRasterLayer[K: SpatialComponent: JsonFormat: ClassTag: Bound
...
+    val clipped: RDD[(SpatialKey, Geometry)] = geometryRDD.clipToGrid(rdd.metadata.layout)
+
+    val addValue = (arr: ArrayBuffer[Geometry], g: Geometry) => g +=: arr
+    val combineArrs = (arr1: ArrayBuffer[Geometry], arr2: ArrayBuffer[Geometry]) =>
+      if (arr1.size > arr2.size) arr2 ++=: arr1 else arr1 ++=: arr2
+
+    val aggregatedRDD: RDD[(SpatialKey, Seq[Geometry])] =
+      rdd.partitioner match {
+        case Some(p) =>
+          clipped.aggregateByKey(ArrayBuffer[Geometry](), p)(addValue, combineArrs).mapValues { _.toSeq }
```
echeipesh: This seems like its trying to avoid calling `.groupByKey`, which would require less setup. Why is that?

Now that I look at it, it also seems like the `.groupByKey` follewed by join can be replaced by a single [`cogroup`](https://spark.apache.org/docs/2.1.1/api/scala/index.html#org.apache.spark.rdd.PairRDDFunctions@cogroup[W1,W2](other1:org.apache.spark.rdd.RDD[(K,W1)],other2:org.apache.spark.rdd.RDD[(K,W2)],partitioner:org.apache.spark.Partitioner):org.apache.spark.rdd.RDD[(K,(Iterable[V],Iterable[W1],Iterable[W2]))])

The left side of the tuple, should only contain a single tile, the right side will contain all the geometries for that tile.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
```diff
@@ -93,6 +93,45 @@ abstract class TiledRasterLayer[K: SpatialComponent: JsonFormat: ClassTag: Bound
...
+        case None =>
+          clipped.aggregateByKey(ArrayBuffer[Geometry]())(addValue, combineArrs).mapValues { _.toSeq }
+      }
+
+    /*
+     * The reason why this produces a () => Option[Partitioner] instead of just an Option[Partitioner]
+     * is due to type erasure when calling this methods from Python.
+     */
```
echeipesh: Could you explain that problem this avoids? I would think that `() => Option[Partitioner]` would become `Function1[Any, Any]` when provided through `py4j` and would face the same problem as `Option[Partitioner]`.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
echeipesh: Also the shuffle here should either use the partitioning strategy, which it can because after clipping its indexed by key or it should use the partitioner of the tile RDD if no strategy is specified.

geopyspark/geotrellis/layer.py
```diff
@@ -1651,25 +1653,63 @@ def sum_series(self, geometries):
         fn = self.srdd.sumSeries
         return self.star_series(geometries, fn)
 
-    def mask(self, geometries):
+    def mask(self,
+             geometries,
+             partition_strategy=None,
+             options=RasterizerOptions()):
```
echeipesh: Really cool that we get to reuse this plumbing here without extra work.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
jbouffard: The keys of the `RDD` are changing, and it's generally not recommended to have `preservesPartitioning` be `true` if that's the case.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
jbouffard: That's apparently not the case. When `partitionStrategy` is given as a `None` from Python it will be turned into a `null` that can be interpreted as either a `PartitionStrategy` or as an `Option`. What was happening was that instead of going `TiledRasterLayer.mask` -> `SpatialTiledRasterLayer.mask` it was going right to the `SpatialTiledRasterLayer.mask` and getting a match error on `null`.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
jbouffard: That's a good point. I noticed that we do use this convention a lot throughout our code bases.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
jbouffard: My reasoning for doing an `aggregateByKey` was to avoid some of the cost of the shuffle brought on by a `groupByKey`. The reason for why I'd have it use the partitioner of the

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TemporalTiledRasterLayer.scala
echeipesh: Ah, right. Thats correct thats concern in general, specifically that when performing `.mapPartitions(???, preservesPartitioning=true)` you could change the value of the key such that it would no longer map to the partition on which it resides. However, because in this case the RDD is result of the join on the keys directly above and the keys are not changed in the `mapPartitions` we're totally safe. Essentially `k == tk` is always true.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
echeipesh: Gotcha.
Both [`groupByKey`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L498-L508) and [`aggregateByKey`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L160-L174) end up calling [`combineByKeyWithClassTag`](https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L71-L86).

The only difference is that `aggregateByKey` gives you a chance to provide an aggregating function that will be used in map-side combine where is `groupByKey` always concatenates all records for a key into a list.
 You would gain a benefit using `aggregateByKey` if the result of the combine would in some way be smaller than the individual parts (for instance merging two tiles into one). But thats not the case here, you're must be maintaining a list.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/TiledRasterLayer.scala
echeipesh: Right so it wouldn't be able to distinguish between:

```scala
def mask(
    aggregatedRDD: RDD[(SpatialKey, Seq[Geometry])],
    partitionStrategy: PartitionStrategy,
    options: Rasterizer.Options
  ): TiledRasterLayer[K]

// and

  def mask(
    wkbRDD: RDD[Array[Byte]],
    partitionStrategy: PartitionStrategy,
    options: Rasterizer.Options
  ): TiledRasterLayer[K] = {
```
because after erasure `RDD[Array[Byte]]` and `RDD[(SpatialKey, Seq[Geometry])]` look exactly the same: `RDD[Any]`. I guess it just picks the one it found first during reflection. So changing  `PartitionStrategy` to `() => Option[Partitioner]` allows it to distinguish between the two as its difference between the two.

I would say that its better to rename one of the `mask` methods rather than introducing a another way to specify the partitioner to avoid this problem.

PR #635, opened 2018-03-06
Remove hbase, cassandra, accumulo dependencies

echeipesh:
GeoPySpark should be able to load these from the class path using the LayerReader SPI.
In the meantime keeping these optional should reduce the number of possible class conflicts and slim down the base uber jar.


jbouffard:
@echeipesh This looks good to me. Could you please squash some of your commits though before merging?

echeipesh:
Yeah, there was a bit of bashing CI on this one.

PR #640, opened 2018-03-13
GPS Wheel and Egg Deployment

jbouffard:
This PR makes it so that Travis will upload the latest `whl` and `egg` files of GPS to S3 every time a branch is merged into master.


jamesmcclain:
I will try to get you a review by tomorrow morning.

PR #641, opened 2018-03-15
to_dict Method for Histogram

jbouffard:
This PR adds the `to_dict` method to the `Histogram` class. The reason for adding this method is because there is currently no easy way to convert a `Histogram` instance into a `dict`. This is needed though if the user wishes to save the `Histogram` as an `Attribute` as `Attributes` can only write values that are `dict`s.


jbouffard:
Turns out that `to_dict` already exists. My bad

PR #643, opened 2018-03-19
SpaceTimePartitionStrategy

### LINKED ISSUE #642: Expose SpaceTimePartitionStrategy
Currently Geopyspark supports a spatialPartitionStrategy, but this does not work for spacetime layers.
Geotrellis does support this, so it would be nice if it is exposed in GeoPyspark.

jbouffard:
This PR adds the `SpaceTimePartitionStrategy` to to the list of `PartitionStrategies`. This will allow for better partitioning of `SPACETIME` layers. The `SpaceTimePartitioner` will try to sort each element of the layer based on its spatial location and its time. Temporal sorting can be controlled both by a unit of time (ie. `DAYS`, `MONTHS`, etc) and by a resolution (ie. 1 day, 3 weeks, etc).

This PR resolves #642

### DIFF
--- a/geopyspark/geotrellis/__init__.py
+++ b/geopyspark/geotrellis/__init__.py
@@ -526,8 +526,15 @@ class SpatialPartitionStrategy(namedtuple("SpatialPartitionStrategy", "num_parti
             partitioning. Default is, ``None``. If ``None`` the resulting layer will have
             a ``HashPartitioner`` with the number of partitions being either the same
             as the input layer's, or a number computed by the method.
-        bits (int, optional): How many bits the resulting ``Key Index`` of a ``Tile`` should
-            be shifted to the right. Default is, ``8``.
+        bits (int, optional): Helps determine how much data should be placed in each
+            partition. Default is, ``8``.
+
+            GeoPySpark uses a Z-order curve to determine how values within the layer should
+            be grouped. This is done by first finding the ``Key Index`` of a value and then performing a
+            bitwise right shift on the resulting index. From the remaining bits, a partition is selected
+            such that those indexes with the same remaining bits will be in the same partition.
+            Therefore, as the number of bits shifted to the right increases, so then too does the
+            group sizes.
 
     Returns:
         :class:`~geopyspark.geotrellis.SpatialPartitionStrategy`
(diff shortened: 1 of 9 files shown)


geopyspark/geotrellis/__init__.py
```diff
@@ -539,6 +539,48 @@ def __new__(cls, num_partitions=None, bits=8):
...
+
+            This value can either be an ``int`` or a string representation of an ``int``.
+
+    Returns:
+        :class:`~geopyspark.geotrellis.SpaceTimePartitionStrategy`
+    """
+
+    __slots__ = []
+
+    def __new__(cls, time_unit, num_partitions=None, bits=8, time_resolution=None):
```
jbouffard: One thing that is worth noting here is that the order of the parameters for `SpaceTimePartitionStrategy` are in a slightly different order than that of `SpatialPartitionStrategy`. This is because `time_unit` is the only required parameter whereas all of the others are optional, and it makes more sense to me to have the required parameter be first.

jdries:
This looks as expected.
For me, the most difficult parameter to use is 'bits' and it's also not entirely clear to me how it relates to num_partitions. Of course, this parameter also exists in plain geotrellis, so it makes sense to also expose it here.

jbouffard:
@jdries The `bits` parameter is kind of confusing, and it's something we should explain better in the documentation.

In order to determine what partition an element belongs to, we first calculate its `z-value` using a [Z-order curve](https://en.wikipedia.org/wiki/Z-order_curve) which in this case is the interleaving of the binary values of the `col`, `row`, and `instant` of the `Tile`. This `z-value` represents an area that is unique to those values. However, this is a problem as we'd like to group values who are close to each other both spatially and temporally in the same partition, and it's difficult to achieve this when all of the `z-value`s are unique. Thus, we perform a right shift on the resulting `z-value` https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialPartitioner.scala#L21

By dropping N number of bits from each `z-value`, we can create larger groupings of values since those elements that are near each other will have their remaining bits in common. Therefore, the larger `bits` is, the larger the groupings of values will be. `bits` controls how many elements will (ideally) be in each partition.

The shifted `z-index` is used to assign its associated value to a partition based on `num_partitions` by doing `z-index % num_partitions`. Which will try to distribute groupings evenly between the partitions.

I hope I was able to answer your question, but please let me know if it wasn't clear. @echeipesh Do you have anything you'd like to add?

geopyspark/geotrellis/__init__.py
```diff
@@ -539,6 +539,48 @@ def __new__(cls, num_partitions=None, bits=8):
...
+
+    Args:
+        time_unit (str or :class:`~geopyspark.geotrellis.constants.TimeUnit`): Which time
+            unit should be used when saving spatial-temporal data. This controls the resolution of
+            each index. Meaning, what time intervals are used to seperate each record.
+        num_partitions (int, optional): The number of partitions that should be used during
+            partitioning. Default is, ``None``. If ``None`` the resulting layer will have
+            a ``HashPartitioner`` with the number of partitions being either the same
+            as the input layer's, or a number computed by the method.
+        bits (int, optional): How many bits the resulting ``Key Index`` of a ``Tile`` should
```
jpolchlo: This is a pretty inside-baseball kind of parameter.  It's crying out for more and better documentation.

jbouffard: Yeah, now that I know more about I'll add a bit more to the docs to make things a little clearer.

jbouffard:
@jpolchlo I updated the docstring. How does it look now?

jpolchlo:
I feel as if there are more words there, but I'm only getting a slightly better understanding of what the parameter is there for.  I'm a neophyte in regards to spatial partitioning, but not so dumb that I shouldn't be able figure it out with a few good sentences.  Is there a way to describe why I would change the `bits` parameter?  What would the tradeoffs be?  Can you impart any intuition about this?  (Viz. the connection between the z-curve and the quadtree?)

jamesmcclain:
> the connection between the z-curve and the quadtree?

I am typing this without looking at the underlying code in question, but typically you would expect one bit per spatial dimension per division.  In the 2D case, the number of bits divided by two is the number of divisions (levels of the quadtree) that are payed-attention-to.

jbouffard:
I'm having some trouble explaining what `bits` does in a concise manner, so I'm thinking about just merging this PR and then make an issue for it. Would that be okay with you @jpolchlo?

It might also be good to make the same issue in GT since I have to yet find any documentation there that explains what `bits` does well.

jpolchlo:
I'm ok with that.  Make an issue!

PR #657, opened 2018-05-16
README documentation cleanup

CloudNiner:
## Overview

This PR attempts to restructure the project README according to the following goals:
- Clarify the requirements and installation setup for new users
  - Simplify installation notes by providing only two user install options, either pip or from source, which is pretty standard across python projects
- Improve documentation for GPS developers
  - By grouping all information related to development into a single top level heading, rather than spread across multiple top-level headings, e.g. dev installation instructions in the installation section, which should generally be focused on user installations.
- Improve access to ReadTheDocs documentation

~~This PR should be considered a first draft effort. There are a few TODOs that I'd like to get input from @jbouffard about, and the sections could be restructured if we think the above goals could be met in other ways.~~ Update: TODOs should all be removed.

The diff is pretty messy, it's likely best to review the changes by looking at the rendered view of the new document.

## Notes

This PR also updates the build instructions in the Contributing docs to reference the README, since those appeared to be out of date.

I don't feel strongly about keeping the Example at the very top. Some projects do this, others start with Requirements + Installation and then show the Example. I suppose Example should stay at top if we primarily want to highlight the project's utility, otherwise swap if we think the utility is clear enough from a short intro/descriptive sentence that we can jump right to setup instructions, pushing the example down.

I signed the Eclipse CLA and commits appear verified, so I believe I have all the contributing prereqs done. Hopefully I didn't miss anything, LMK if I need to address anything else for that.


CloudNiner:
We had also originally talked about adding a "Getting Started" section to the ReadTheDocs documentation. Rather than writing that up next, it may be worth just moving the Requirements / Installation notes changed here to the docs and just specifically linking to those sections from the README, rather than attempting to continue to maintain duplicate instruction sets in two places.

README.rst
```diff
@@ -172,21 +141,11 @@ Note:
...
-sub-package. The second command will install GeoPySpark in "editable" mode.
-Meaning any changes to the source files will also appear in your system
-installation.
-
 Installing to a Virtual Environment
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+TODO: Ask Jake if he's seen issues with this. I had no problem using the standard pip installation
+instructions in a virtualenv. If no problems, perhaps we remove for now.
```
jpolchlo: I'll say that installing into a virtualenv gave me weird problems when trying to get this thing running in Jupyter.  IPython allows `import geopyspark`, but Jupyter does not.  The `sys.path` is lacking a number of .eggs that are needed (shapely, protobuf, and a couple others).  We'd probably want to figure out if this is a reproducible bug, and if so, provide advice on how to fix it.  Jupyter is expected to be a common use.

CloudNiner: Gotcha. I just did `pip install jupyter && jupyter notebook` then loaded up each of the demos in the `notebook-demos` folder without issue. I did have to `pip install rasterio` for the "Reading in Sentinel data" demo and `pip install pyproj` for the "Layers" demo, but saw no other errors running each of the demos. None of the demos used protobuf as far as I saw, but a few used shapely, which didn't have any trouble.

This was with gps 0.4.0 installed via `pip` using a "recent" version of virtualenvwrapper installed within a month or so using Python 3.6.4. 

While not explicitly stated anywhere, the way the section is written alludes to this issue only cropping up when GPS is _installed from source_ into a virtualenv. I attempted to test this, but ran into what looks like an unrelated GT compiler error when running `make install` from the current master branch:
```
[info] Compiling 73 Scala sources to /Users/andrew/src/geopyspark/geopyspark-backend/geotrellis/target/scala-2.11/classes...
[error] /Users/andrew/src/geopyspark/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala:592: type mismatch;
[error]  found   : Long
[error]  required: Int
[error]         case ps: PartitionStrategy => ps.producePartitioner(math.max(gb.size / 512, 1))
[error]                                                                              ^
[error] /Users/andrew/src/geopyspark/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala:636: type mismatch;
[error]  found   : Long
[error]  required: Int
[error]         case ps: PartitionStrategy => ps.producePartitioner(math.max(gb.size / 512, 1))
[error]                                                                              ^
```

So I'll have to investigate that further a bit later.

jpolchlo: Yeah, interesting.  I'll clarify that I ran into my troubles when installing from source as well.

¯\_(ツ)_/¯

jbouffard: I've personally never had issues installing GPS either from source or via pip in a virutalenv.

@jpolchlo That's pretty weird behavior. The only thing I can think of that might cause that is if GPS, Jupyter, and all of the other libraries weren't installed in the same virtualenv.

@CloudNiner Yeah, there is a potential issue when installing into a virtualenv from source when the user does `pip install -e .`. As Python will then look at the path of that directory for GPS, and if that folder is ever moved it'll cause errors.

README.rst
```diff
@@ -1,19 +1,21 @@
...
-***********
+**********
+
 .. image:: https://travis-ci.org/locationtech-labs/geopyspark.svg?branch=master
    :target: https://travis-ci.org/locationtech-labs/geopyspark
 
 .. image:: https://readthedocs.org/projects/geopyspark/badge/?version=latest
    :target: https://geopyspark.readthedocs.io/en/latest/?badge=latest
 
+Documentation: https://geopyspark.readthedocs.io
```
jbouffard: Do we need a link to the documentation right below the badge that links to the documentation?

README.rst
```diff
@@ -1,19 +1,21 @@
 GeoPySpark
-***********
+**********
+
 .. image:: https://travis-ci.org/locationtech-labs/geopyspark.svg?branch=master
    :target: https://travis-ci.org/locationtech-labs/geopyspark
 
 .. image:: https://readthedocs.org/projects/geopyspark/badge/?version=latest
    :target: https://geopyspark.readthedocs.io/en/latest/?badge=latest
```
jbouffard: Since we're moving down the `Contact and Support` section, it might make sense to add a Gitter badge here to make it easier for people to get there without having to scroll down the page.

README.rst
CloudNiner: Only anecdotal, but Jenny, Ross and I were all unaware that there was a readthedocs page. I think I knew that the docs badge was also a link, but my brain skipped over the fact that I could click that. Given that the documentation is useful, I personally still think a direct link is worth the space as a callout. Happy to revert though.

README.rst
CloudNiner: Good point and happy to make that change. Alternatively, should we move the `Contact and Support` section farther up? Perhaps Example -> Contact/Support -> Requirements + Setup instead? Or do both?

README.rst
CloudNiner: Ok. Is the `pip install -e .` an edge case, or is only useful when developing GPS? I think if the former, we should remove it or fixup the wording to clarify when it would be necessary. If the latter, we can move this section down into the `Developing GeoPySpark -> Installing for Developers` section.

README.rst
jbouffard: Ah, I see. In that case I'm okay with making the link explicit then. In that case I think we should make an explicit Gitter link as well since the badge alone might not be enough.

README.rst
jbouffard: I think having the `Contact and Support` section further down in the document makes sense. As it seems to interrupt the flow of introducing GPS and its setup if it's earlier in the document.

README.rst
jbouffard: I think that `pip install -e .` would only be useful for those developing GPS. I can't see much benefit for someone just looking to use it. That issue isn't specific to GPS either, so I'm not sure if there's really anything we should mention if that's the way they want to set things up.

CloudNiner: Sounds good. I moved it to a short subsection in the Installation for Developers section. Should be visible, but out of the way.

CloudNiner:
I pushed updates for the changes we discussed just now, which include:
- Added gitter badge
- Added links section for Documentation + Gitter links at the top
- Removed getting started example from ReadTheDocs documentation
- Moved the notes about the `PYTHONPATH + pip install -e .` installation to "Installation for Developers"
- Removed all remaining TODOs, Justin is going to address the Geonotebook section separately

PR #571, opened 2017-12-06
[WiP] Add Rasterio Tile-Reading Capability

jamesmcclain:

PR #661, opened 2018-05-24
Add RasterIO tile reading capability

echeipesh:
This is revival of: https://github.com/locationtech-labs/geopyspark/pull/571


geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+
+# On driver
+if "GDAL_DATA" in os.environ:
+    __GDAL_DATA = os.environ["GDAL_DATA"]
+else:
+    __GDAL_DATA = None
+
+
+def dataset_to_proj4(dataset):
+    from osgeo import osr
```
jbouffard: Similar to the `rasterio` import as we don't require `osgeo` to be installed to use GPS
```python
try:
    from osgeo import osr
except ImportError:
    raise ImportError("osgeo must be installed in order to use the dataset_to_proj4 function")
```

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+        xcols: The desired tile width.
+        ycols: The desired tile height.
+        bands: The bands from which tiles should be produced.  An array of integers.
+
+    Returns:
+        An array of dictionaries.  The tile can be found in a numpy array under the key ``data``.
+    """
+
+    metadata = []
+    for band in bands:
```
jbouffard: This won't produce the expected result. It'll split up a MultibandGeoTiff into separate bands and then tile them.

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+        band = new_metadatum['band']
+        new_metadatum['data'] = dataset.read(band, window=metadatum['window'])
+        new_metadatum.pop('window')
+        new_metadatum.pop('uri')
+        new_metadatum.pop('band')
+
+    return new_metadatum
+
+
+def uri_to_pretiles(uri, dataset_to_proj4=dataset_to_proj4, xcols=512, ycols=512, bands=[1]):
```
jbouffard: The `bands` parameter should be set to `None` by default. Because right now it'll only read the first band of file regardless of how many there are.

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+    """Given a URI, this method uses rasterio to generate series of tiles of the desired dimensions.
+
+    Args:
```
jbouffard: To match the docstrings in other parts of the library, these should be expanded to include the types of the parameters. Here is the `geotiff.get` docstring for example https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/geotiff.py#L39-L80

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+        ycols: The desired tile height.
+        bands: The bands from which tiles should be produced.  An array of integers.
+
+    Returns:
+        An array of dictionaries.  The tile can be found in a numpy array under the key ``data``.
+    """
+
+    metadata = []
+    for band in bands:
+        metadata = metadata + __get_metadata(uri, dataset_to_proj4, xcols, ycols, band)
```
jbouffard: This doesn't seem like an efficient way of collecting the `Metadata`, as the resulting list could become very long. A list comprehension would be better here.

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+    return [__get_data(metadatum) for metadatum in metadata]
+
+
+def pretile_to_tile(pretile, no_data_arg=None):
+    """Given a pretile (as produced by ``uri_to_pretiles``), produce a projected extent and a tile.
+
+    Args:
```
jbouffard: Same comment on the `Args` section as the one above.

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
...
+
+    return metadata
+
+
+def __get_data(metadatum):
+    new_metadatum = metadatum.copy()
+
+    with rasterio.open(metadatum['uri']) as dataset:
+        band = new_metadatum['band']
+        new_metadatum['data'] = dataset.read(band, window=metadatum['window'])
```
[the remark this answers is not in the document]
jbouffard: Actually, we should probably just create a GPS `Tile` instance here since we already have the data needed.

setup.py
```diff
@@ -23,6 +23,7 @@
     ],
     packages=[
         'geopyspark',
+        'geopyspark.command',
```
jbouffard: I think that this branch might have to be rebassed, because I'm not sure why else this addition would be here. Also, `geopyspark.rasterio` needs to be added to this list.

geopyspark/rasterio/read.py
```diff
@@ -0,0 +1,123 @@
+import os
+import math
+import rasterio
```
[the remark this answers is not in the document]
echeipesh: There is a good argument to make tests contingent on rasterio bing present too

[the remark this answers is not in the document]
jbouffard: That's true, but I feel that only people developing GPS will be running the tests. For those who just want to use GPS, rasterio is an unneeded requirement. We also do have rasterio listed in the `requirements.txt` https://github.com/locationtech-labs/geopyspark/blob/master/requirements.txt#L4

.vscode/tags
```diff
@@ -0,0 +1,2934 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
jbouffard: Is this file needed? I'm not sure what it does.

geopyspark/geotrellis/rasterio.py
```diff
@@ -0,0 +1,91 @@
...
+        from osgeo import osr
+    except ImportError:
+        raise ImportError("osgeo must be installed in order to use the crs_to_proj4 function")
+
+    srs = osr.SpatialReference()
+    srs.ImportFromWkt(crs.wkt)
+    proj4 = srs.ExportToProj4()
+    return proj4
+
+def read_windows(uri, xcols=512, ycols=512, bands=[1], crs_to_proj4=crs_to_proj4):
```
jbouffard: Is there a benefit for having the default number of cols and rows to be 512? Because in `geotiff.get` we have it set to 256 https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/constants.py#L17

jbouffard: I think it'd be better for have `bands=None` rather than `[1]`. Because if you're reading in a MultibandTile than you'll just get the first band by default.

geopyspark/geotrellis/rasterio.py
```diff
@@ -0,0 +1,91 @@
...
+            bottom = bounds.top + (bounds.bottom - bounds.top)*(float(row_stop)/height)
+            top = bounds.top + (bounds.bottom - bounds.top)*(float(row_start)/height)
+            extent = gps.Extent(left, bottom, right, top)
+            projected_extent = gps.ProjectedExtent(extent=extent, proj4=proj4)
+
+            data = dataset.read(bands, window=window)
+            tile = gps.Tile.from_numpy_array(data, no_data_value=nodata)
+            yield (projected_extent, tile)
+
+def get(uri):
```
jbouffard: `get` should have the same parameters as `read_windows` since it's just a wrapper of that function. It might be worth having `read_windows` be private since it doesn't deal with Spark at all.

geopyspark/geotrellis/rasterio.py
```diff
@@ -0,0 +1,91 @@
...
+        xcols: The desired tile width.
+        ycols: The desired tile height.
+        bands: The bands from which windows should be produced.  An array of integers.
+        crs_to_proj4: A that takes a :class:`rasterio.crs.CRS` and returns a Proj4 string.
+
+    Returns:
+        Generator of (:class:`~geopyspark.geotrellis.ProjectedExtent`, :class:`~geopyspark.geotrellis.Tile`)
+    """
+
+    if ("GDAL_DATA" not in os.environ) and (_GDAL_DATA != None):
```
jbouffard: I'm still not sure how `_GDAL_DATA` is set when it's not in the environment.

geopyspark/geotrellis/rasterio.py
```diff
@@ -0,0 +1,91 @@
...
+
+    srs = osr.SpatialReference()
+    srs.ImportFromWkt(crs.wkt)
+    proj4 = srs.ExportToProj4()
+    return proj4
+
+def read_windows(uri, xcols=512, ycols=512, bands=[1], crs_to_proj4=crs_to_proj4):
+    """Given a URI, this method uses rasterio to generate series of windows of the desired dimensions.
+
+    Args:
```
jbouffard: The `Args` section of the docs should be expanded so there similar to the other docs in the library. For example, https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/geotiff.py#L32-L76

geopyspark/geotrellis/rasterio.py
echeipesh: Not sure the mechanics here, this is something I transcribed from James work.
This is intended to handle the case where `GDAL_DATA` is set on master but not the workers.
Presumably the only way it works if `_GDAL_DATA` is captured as a closures in `PipelinedRDD`. I'm honestly not sure how or if that process actually happens here.

jbouffard: I see. I'm fine with keeping it in for now just to make sure everything works. We can look into removing it later.

PR #668, opened 2018-06-26
RasterFrames Methods

jbouffard:
This PR adds two methods to `TiledRasterLayer`: `to_rasterframe` and `from_rasterframe`. These methods will convert a `TiledRasterLayer` into a `RasterLayer` and a `RasterLayer` into a `TiledRasterLayer`, respectively.

**Note:** This PR should not be merged until deployment issues with `pyrasterframes` have been resolved.

### DIFF
--- a/geopyspark/geotrellis/layer.py
+++ b/geopyspark/geotrellis/layer.py
@@ -1028,6 +1028,68 @@ class TiledRasterLayer(CachableLayer, TileLayer):
 
         return cls(layer_type, srdd)
 
+    @classmethod
+    def from_rasterframe(cls, rasterframe, zoom_level=None):
+        """Creates a ``TiledRasterLayer from a ``pyrasterframes.RasterFrame``.
+
+        Note:
+            ``pyrasterframes`` needs to initialized via the ``.withRasterFrames()``
+            extension method on the active ``SparkSession`` object in order to
+            use this method.
+
+        Args:
+            rasterframe (pyrasterframes.RasterFrame): The target ``RasterFrame`` that
+                will be converted into a ``TiledRasterLayer``.
+            zoom_level(int, optional): The ``zoom_level`` the resulting `TiledRasterLayer` should
+                have. If ``None``, then the returned layer's ``zoom_level`` will be ``None``.
+
+        Returns:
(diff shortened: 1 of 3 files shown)


geopyspark/geotrellis/layer.py
```diff
@@ -1028,6 +1028,68 @@ def from_numpy_rdd(cls, layer_type, numpy_rdd, metadata, zoom_level=None):
 
         return cls(layer_type, srdd)
 
+    @classmethod
+    def from_rasterframe(cls, rasterframe, zoom_level=None):
+        """Creates a ``TiledRasterLayer from a ``pyrasterframes.RasterFrame``.
+
+        Note:
+            ``pyrasterframes`` must be installed, its associated jar
+            has to be added to the Spark path, and ``RasterFrames`` needs
```
jpolchlo: ``RasterFrames`` needs to be enabled?  Do you mean to say that ``.withRasterFrames()`` needs to be called on the spark session object?

geopyspark/geotrellis/layer.py
```diff
@@ -1028,6 +1028,68 @@ def from_numpy_rdd(cls, layer_type, numpy_rdd, metadata, zoom_level=None):
...
+            layer_type = LayerType.SPATIAL
+
+        return cls(layer_type, srdd)
+
+    def to_rasterframe(self, num_bands):
+        """Converts a ``TiledRasterLayer`` to a ``pyrasterframes.RasterFrame``.
+
+        Note:
+            ``pyrasterframes`` must be installed, its associated jar
+            has to be added to the Spark path, and ``RasterFrames`` needs
```
jpolchlo: Same comment as above.

geopyspark/geotrellis/layer.py
jbouffard: Yeah, should I make that more clear in the docs?

jpolchlo: You could easily say "``RasterFrames`` needs to be initialized via the ``.withRasterFrames()`` extension method on the active SparkSession object."

jbouffard: That's true. I'll make that change now

PR #671, opened 2018-07-13
Add interface for key/layout computations

jpolchlo:
It is useful at times to be able to relate geometries to grid cells in some layout, and keys to geometries in space wrt the layout.  This PR adds that feature, allowing geometries and extents, represented using python-native `gps.Extent`s and shapely geometries to be converted into `gps.SpatialKeys` and `gps.SpatialKey`s to be converted into `gps.Extent`s.

### DIFF
--- a/docs/docs/geopyspark.geotrellis.key_conversion.rst
+++ b/docs/docs/geopyspark.geotrellis.key_conversion.rst
@@ -1,0 +1,6 @@
+geopyspark.geotrellis.key_conversion module
+=====================================
+
+.. automodule:: geopyspark.geotrellis.key_conversion
+   :members:
+   :inherited-members:
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/util/GeometryUtils.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/util/GeometryUtils.scala
@@ -1,0 +1,10 @@
+package geopyspark.geotrellis.util
+
+import geotrellis.vector.Geometry
+import geotrellis.vector.io.wkb.WKB
+
+object GeometryUtil {
+
+  def wkbToScalaGeometry(wkb: Array[Byte]): Geometry = WKB.read(wkb)
+
+}
--- a/geopyspark/geotrellis/key_conversion.py
+++ b/geopyspark/geotrellis/key_conversion.py
@@ -1,0 +1,119 @@
+'''
+A module to provide facilites for converting between layout keys and spatial objects.  These facilities aim to bridge between the various layouts, ``SpatialKey``s, and geometry.
+'''
+
+from math import ceil
+
+import geopyspark as gps
+
+__all__ = ["KeyTransform"]
+
+class KeyTransform(object):
+    """Provides functions to move from keys to geometry and vice-versa.
+
+    Tile Layers have an underlying RDD which is keyed by either :class:`geopyspark.geotrellis.SpatialKey` or
+    :class:`geopyspark.geotrellis.SpaceTimeKey`.  Each key represents a region in space, depending on a choice of layout.
+    In order to enable the conversion of keys to regions, and of geometry to keys, the ``KeyTransform`` class is provided.
+    This class is constructed with a layout, which is either ``GlobalLayout``, ``LocalLayout``, or a ``LayoutDefinition``.
+    Global layouts use power-of-two pyramids over the world extent, while local layouts operate over a defined extent and
+    cellsize.
+
+    NOTE: LocalLayouts will encompass the requested extent, but the final layout may include ``SpatialKey``s which only
+    partially cover the requested extent.  The upper-left corner of the resulting layout will match the requested extent,
+    but the right and bottom edges may be beyond the boundaries of the requested extent.
+
+    Args:
+        layout(:class:`geopyspark.geotrellis.GlobalLayout` or :class:`geopyspark.geotrellis.LocalLayout`
+            or :class:`geopyspark.geotrellis.LayoutDefinition`): a definition of the layout scheme defining the key structure.
+        crs (str or int): Used only when `layout` is :class:`geopyspark.geotrellis.GlobalLayout`.  Target CRS of reprojection.
+            Either EPSG code, well-known name, or a PROJ.4 string
+        extent (:class:`geopyspark.geotrellis.Extent`): Used only for ``LocalLayout``s.  The area of interest.
+        cellsize (tup of (float, float)): Used only for ``LocalLayout``s.  The (width, height) in extent units of a pixel.
+            Cannot be specified simultaneously with ``dimensions``.
+        dimensions (tup of (int, int)): Used only for ``LocalLayout``s.  The number of (columns, rows) of pixels over the
+            entire extent.  Cannot be specified simultaneously with ``cellsize``.
+    """
+
+    def __init__(self, layout, crs=None, extent=None, cellsize=None, dimensions=None):
+        self.__jvm = gps.get_spark_context()._gateway.jvm
+
+        if isinstance(layout, gps.LocalLayout):
+            if not extent:
+                raise ValueError("Must specify an extent when using LocalLayout")
+
+            if dimensions and not cellsize:
+                cellsize = ((extent.xmax - extent.xmin)/dimensions[0], (extent.ymax - extent.ymin)/dimensions[1])
+                dimensions = None
+
+            if cellsize and not dimensions:
+                tilewidth = layout.tile_cols * cellsize[0]
+                tileheight = layout.tile_rows * cellsize[1]
+                rows = ceil((extent.xmax - extent.xmin) / tilewidth)
+                cols = ceil((extent.ymax - extent.ymin) / tileheight)
+                extent = gps.Extent(extent.xmin, extent.ymax - rows * tileheight, extent.xmin + cols * tilewidth, extent.ymax)
+                tl = gps.TileLayout(cols, rows, layout.tile_cols, layout.tile_rows)
+            else:
+                raise ValueError("For LocalLayout, must specify exactly one: cellsize or dimension")
+        elif isinstance(layout, gps.GlobalLayout):
+            from pyproj import Proj, transform
+
+            if not crs:
+                raise ValueError("Must specify a crs when using GlobalLayout")
+
+            if isinstance(crs, int):
+                crs = "{}".format(crs)
+
+            gtcrs = self.__jvm.geopyspark.geotrellis.TileLayer.getCRS(crs).get()
+
+            if gtcrs.epsgCode().isDefined() and gtcrs.epsgCode().get() == 3857:
+                extent = gps.Extent(-20037508.342789244, -20037508.342789244, 20037508.342789244, 20037508.342789244)
+            elif gtcrs.epsgCode().isDefined() and gtcrs.epsgCode().get() == 4326:
+                extent = gps.Extent(-180.0, -89.99999, 179.99999, 89.99999)
+            else:
+                llex = gps.Extent(-180.0, -89.99999, 179.99999, 89.99999)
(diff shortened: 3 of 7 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/util/GeometryUtils.scala
```diff
@@ -0,0 +1,10 @@
+package geopyspark.geotrellis.util
+
+import geotrellis.vector.Geometry
+import geotrellis.vector.io.wkb.WKB
+
+object GeometryUtil {
+
+  def wkbToScalaGeometry(wkb: Array[Byte]): Geometry = WKB.read(wkb)
```
jbouffard: This isn't something that needs to be done now, but I'd be interested in seeing the performance difference between sending Geometries as `Array[Byte]` vs `String` between Scala/Python.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+                tilewidth = layout.tile_cols * cellsize[0]
+                tileheight = layout.tile_rows * cellsize[1]
+                rows = ceil((extent.xmax - extent.xmin) / tilewidth)
+                cols = ceil((extent.ymax - extent.ymin) / tileheight)
+                extent = gps.Extent(extent.xmin, extent.ymax - rows * tileheight, extent.xmin + cols * tilewidth, extent.ymax)
+                tl = gps.TileLayout(cols, rows, layout.tile_cols, layout.tile_rows)
+            else:
+                raise ValueError("For LocalLayout, must specify exactly one: cellsize or dimension")
+        elif isinstance(layout, gps.GlobalLayout):
+            from pyproj import Proj, transform
```
jbouffard: `pyproj` isn't a required library to use GPS, so there should be a `try catch` here when importing it. Similar to what we did in the `rasterio` module: https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/rasterio.py#L7-L10

jbouffard: An additional `NOTE` should also be made in the docstring stating this requirement when using a `GlobalLayout`.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+            if not crs:
+                raise ValueError("Must specify a crs when using GlobalLayout")
+
+            if isinstance(crs, int):
+                crs = "{}".format(crs)
+
+            gtcrs = self.__jvm.geopyspark.geotrellis.TileLayer.getCRS(crs).get()
+
+            if gtcrs.epsgCode().isDefined() and gtcrs.epsgCode().get() == 3857:
+                extent = gps.Extent(-20037508.342789244, -20037508.342789244, 20037508.342789244, 20037508.342789244)
```
jbouffard: This and the other `Extent`s in this block can be defined as constants.

We could make an Enum class in this module that contains them:

```python
class WorldExtent(Enum):
    WEB_MERCATOR = gps.Extent(-20037508.342789244, -20037508.342789244, 20037508.342789244, 20037508.342789244)
    LATLNG = gps.Extent(-180.0, -89.99999, 179.99999, 89.99999)
```

Or could just assign them in the module itself:

```python
WEB_MERCATOR_WORLD_EXTENT = gps.Extent(-20037508.342789244, -20037508.342789244, 20037508.342789244, 20037508.342789244)
LATLNG_WORLD_EXTENT = gps.Extent(-180.0, -89.99999, 179.99999, 89.99999)
```

We use both forms in GPS, so I'm fine with either implementation.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+                extent = gps.Extent(-180.0, -89.99999, 179.99999, 89.99999)
+            else:
+                llex = gps.Extent(-180.0, -89.99999, 179.99999, 89.99999)
+                proj4str = gtcrs.toProj4String()
+                target = Proj(proj4str)
+                xmin, ymin = target(llex.xmin, llex.ymin)
+                xmax, ymax = target(llex.xmax, llex.ymax)
+                extent = gps.Extent(xmin, ymin, xmax, ymax)
+
+            if not layout.zoom:
```
jbouffard: I think this can be moved up to where the `crs` input is checked. Since there's no reason to do the work of determining the `extent` if `zoom` isn't defined.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+                raise ValueError("Must specify a zoom level")
```
jbouffard: `when using GlobalLayout` should be added to the end of that error message to help avoid confusion.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+        elif isinstance(key, tuple):
+            skey = self.__jvm.geotrellis.spark.SpatialKey(key[0], key[1])
+        elif isinstance(key, int) and len(args) == 1 and isinstance(args[0], int):
+            skey = self.__jvm.geotrellis.spark.SpatialKey(key, args[0])
+        else:
+            raise ValueError("Please supply either gps.SpatialKey, gps.SpaceTimeKey, (int, int), or two ints")
+        ex = self.__layout.mapTransform().apply(skey)
+        return gps.Extent(ex.xmin(), ex.ymin(), ex.xmax(), ex.ymax())
+
+    def extentToKeys(self, extent):
```
jbouffard: Same as the above method in terms of docstrings and name.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+        ex = self.__jvm.geotrellis.vector.Extent(float(extent.xmin), float(extent.ymin), float(extent.xmax), float(extent.ymax))
+        gridbnd = self.__layout.mapTransform().apply(ex)
+        cmin = gridbnd.colMin()
+        cmax = gridbnd.colMax()
+        rmin = gridbnd.rowMin()
+        rmax = gridbnd.rowMax()
+        return (gps.SpatialKey(c, r) for c in range(cmin, cmax + 1) for r in range(rmin, rmax + 1))
+
+    def geometryToKeys(self, geom):
```
jbouffard: This needs a docstring

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+            tl = gps.TileLayout(layout_rows_cols, layout_rows_cols, layout.tile_size, layout.tile_size)
+        elif isinstance(layout, gps.LayoutDefinition):
+            extent = layout.extent
+            tl = layout.tileLayout
+
+        ex = self.__jvm.geotrellis.vector.Extent(float(extent.xmin), float(extent.ymin), float(extent.xmax), float(extent.ymax))
+        tilelayout = self.__jvm.geotrellis.raster.TileLayout(int(tl[0]), int(tl[1]), int(tl[2]), int(tl[3]))
+        self.__layout = self.__jvm.geotrellis.spark.tiling.LayoutDefinition(ex, tilelayout)
+
+    def keyToExtent(self, key, *args):
```
jbouffard: This needs a docstring

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+        if isinstance(key, gps.SpatialKey) or isinstance(key, gps.SpaceTimeKey):
```
jbouffard: This can be rewritten as:

```python
if isinstance(key, (gps.SpatialKey, gps.SpaceTimeKey)):
```

geopyspark/geotrellis/key_conversion.py
jbouffard: Also, Python methods use snake case, so the method should be called, `key_to_extent`

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,157 @@
...
+
+from . import Extent
+import geopyspark as gps
+
+__all__ = ["KeyTransform"]
+
+class KeyTransform(object):
+    """Provides functions to move from keys to geometry and vice-versa.
+
+    Tile Layers have an underlying RDD which is keyed by either :class:`geopyspark.geotrellis.SpatialKey` or
```
jbouffard: When referencing another object in the docstrings, a `~` needs to come before the path to the object (ie ``:class:`~geopyspark.geotrellis.SpatialKey` ``). Otherwise, I don't think it'll link correctly.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,157 @@
+'''
+A module to provide facilites for converting between layout keys and spatial objects.  These facilities aim to bridge between the various layouts, ``SpatialKey``s, and geometry.
```
jbouffard: In rst, if you have a character that isn't a punctuation mark next to highlighted code it'll mess up the formatting. To get around that, a `\` needs to be added before the character. So ` ``SpatialKey``s ` becomes: ` ``SpatialKey``\s `

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,157 @@
...
+        Returns:
+            [:class:`~geopyspark.geotrellis.SpatialKey`]
+        """
+        from shapely.wkb import dumps
+        jts_geom = self.__jvm.geopyspark.geotrellis.util.GeometryUtil.wkbToScalaGeometry(dumps(geom))
+        scala_key_set = self.__layout.mapTransform().keysForGeometry(jts_geom)
+        key_set = self.__jvm.scala.collection.JavaConverters.setAsJavaSetConverter(scala_key_set).asJava()
+        return [gps.SpatialKey(key.col(), key.row()) for key in key_set]
+
+    """The WebMercator (EPSG=3857) world extent"""
```
jbouffard: I think that these constants should be moved out of the class and into the module. Since there's not really a reason for each class instance to have their own constants.

geopyspark/geotrellis/key_conversion.py
```diff
@@ -0,0 +1,157 @@
...
+                cols = ceil((extent.ymax - extent.ymin) / tileheight)
+                extent = gps.Extent(extent.xmin, extent.ymax - rows * tileheight, extent.xmin + cols * tilewidth, extent.ymax)
+                tl = gps.TileLayout(cols, rows, layout.tile_cols, layout.tile_rows)
+            else:
+                raise ValueError("For LocalLayout, must specify exactly one: cellsize or dimension")
+        elif isinstance(layout, gps.GlobalLayout):
+            try:
+                from pyproj import Proj, transform
+            except:
+                raise Exception('pyproj is required for GlobalLayout')
```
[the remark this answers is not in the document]
jpolchlo: For frame of reference, I followed the example set [here](https://github.com/locationtech-labs/geopyspark/blob/master/geopyspark/geotrellis/color.py#L46) when I used Exception.  I can fix both instances to use ImportError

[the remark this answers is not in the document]
jbouffard: Oh, I see. We'll need to fix that as well at some point.

jpolchlo:
Travis errors are unrelated to the changes in this PR.  One is an install error, the rest are linting errors due to `geopyspark/geotrellis/rasterio.py`.  Should be good to merge?

docs/docs/geopyspark.geotrellis.key_conversion.rst
```diff
@@ -0,0 +1,6 @@
+geopyspark.geotrellis.key_conversion module
```
jbouffard: Could have this listed in `geopyspark.geotrellis.rst` under the `Modules` tree (it's at the bottom of the file). That way that document will appear in docs directory on the left side of the page.

PR #673, opened 2018-08-02
update_layer Function for catalog

### LINKED ISSUE #522: Add update_layer Method for catalog
We should add an `update_layer` method to the `catalog` module. This will update a preexisting layer with what ever new data the user gives it. This shouldn't be difficult to implement as this method is just a wrapper for the `update` GeoTrellis method https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/io/LayerWriter.scala#L95

jbouffard:
This PR adds the `update` function to the `catalog` module. This will allow users to overwrite a per-existing layer with a new one.

This PR resolves #522


jbouffard:
@echeipesh @jpolchlo If one of you guys have the time, could you please review this PR?

PR #678, opened 2018-08-16
Write all pyramids

phloem7:
Writes all pyramid layers through Pyramid.write()


geopyspark/geotrellis/layer.py
```diff
@@ -33,9 +33,11 @@
                                    HashPartitionStrategy,
                                    SpatialPartitionStrategy,
                                    SpaceTimePartitionStrategy,
-                                   RasterizerOptions)
+                                   RasterizerOptions,
+                                   catalog)
```
jbouffard: Instead of importing `catalog` could you try importing `write` instead? It's not really clear to me what's going on with those two jobs, so maybe that'll fix it?

jbouffard:
@phloem7 I think this PR is good :+1: Could you just squash some of your commits? Then it'll be ready to be merged.

PR #691, opened 2018-10-19
to_geotiff_rdd Will Now Produce GeoTiffs that Can Be Used as COGs

jbouffard:
This PR improves the `to_geotiff_rdd` method so that the GeoTiffs produced can be used as COGs. In order to achieve this, two changes were made. The first was to change the default `storage_method` from `STRIPED` to `TILED`. The other change was the addition of two new parameters to `to_geotiff_rdd`: `resample_method` and `decimations`. These two parameters are used to create the internal overviews of the outputted GeoTiffs.

*Note:* This PR may break any preexisting code that uses the `to_geotiff_rdd` method, as it changes the order of the arguments to that method.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
@@ -411,20 +411,29 @@ class SpatialTiledRasterLayer(
 
   def toGeoTiffRDD(
     tags: Tags,
-    geotiffOptions: GeoTiffOptions
+    resampleMethod: ResampleMethod,
+    decimations: List[Int],
+    geoTiffOptions: GeoTiffOptions
   ): JavaRDD[Array[Byte]] = {
-    val mapTransform = MapKeyTransform(
-      rdd.metadata.layout.extent,
-      rdd.metadata.layout.layoutCols,
-      rdd.metadata.layout.layoutRows)
+    val mapTransform =
+      MapKeyTransform(
+        rdd.metadata.layout.extent,
+        rdd.metadata.layout.layoutCols,
+        rdd.metadata.layout.layoutRows
+      )
 
     val crs = rdd.metadata.crs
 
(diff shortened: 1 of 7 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/SpatialTiledRasterLayer.scala
```diff
@@ -411,20 +411,30 @@ class SpatialTiledRasterLayer(
...
-      val transKey = ProjectedExtent(mapTransform(x._1), crs)
+    val geotiffRDD =
+      rdd.map { case (k, v) =>
+        val transKey = ProjectedExtent(mapTransform(k), crs)
 
-      (x._1, MultibandGeoTiff(x._2, transKey.extent, transKey.crs, tags, geotiffOptions).toByteArray)
-    }
+        val geoTiff = {
+          val gtiff: GeoTiff[MultibandTile] = MultibandGeoTiff(v, transKey.extent, transKey.crs, tags, geoTiffOptions)
+          gtiff.withOverviews(resampleMethod, decimations)
```
echeipesh: Not really worth a change but it would be my idiomatic and less text just to smash these together.

```scala
MultibandGeoTiff(v, transKey.extent, transKey.crs, tags, geoTiffOptions)
    .withOverviews(resampleMethod, decimations)
```

geopyspark/geotrellis/layer.py
```diff
@@ -454,9 +469,11 @@ def to_png_rdd(self, color_map):
         return create_python_rdd(result, ser)
 
     def to_geotiff_rdd(self,
-                       storage_method=StorageMethod.STRIPED,
+                       storage_method=StorageMethod.TILED,
                        rows_per_strip=None,
                        tile_dimensions=(256, 256),
+                       resample_method=ResampleMethod.NEAREST_NEIGHBOR,
+                       decimations=[2, 4, 8, 16],
```
echeipesh: @jbouffard I feel like these should be off by default. This is extra work and storage requirement that may or may not be helpful and won't always be expected.

PR #693, opened 2018-10-29
read and read_to_layout Methods

jbouffard:
This PR adds the `read` and `read_to_layout` class methods to `RasterLayer` and `TiledRasterLayer` respectively. These two methods will utilize `geotrellis.contrib.vlm` in order to read in data and create these layers.

**Note:** The `read_to_layout` method currently does not work due to sampling issues that occur around the edges of the tiles.


geopyspark/geotrellis/layer.py
```diff
@@ -1010,9 +1075,80 @@ def __init__(self, layer_type, srdd):
         self.layer_metadata = Metadata.from_dict(json.loads(self.srdd.layerMetadata()))
         self.zoom_level = self.srdd.getZoom()
 
+    @classmethod
+    def read_to_layout(cls,
```
echeipesh: Is `read_to_layout` final choice for the name? I recall it was said that existence is layout is implied by the class name in `TiledRasterLayer.read_to_layout` but not sure if that argument for calling out in difference made a bigger sway.

jbouffard: That wasn't a final choice, just something we initially thought of. I agree that the `layout` portion is implicitly understood, and I have no issue changing the name of the method to `read`.

echeipesh: I don't have a hard preference in mind. What do you think is best for the API ?

jbouffard: I think making it `read` would be good. Just so that the API is more consistent.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -0,0 +1,158 @@
...
+    rdd: RDD[String],
+    layoutType: LayoutType,
+    targetCRS: String,
+    resampleMethod: ResampleMethod,
+    readMethod: String
+  ): SpatialTiledRasterLayer = {
+    // TODO: These are the things that still need to be done:
+    // 1. Support TemporalTiledRasterLayer (ie. generic K)
+    // 2. Use the partitionStrategy parameter
+
```
echeipesh: Needs an extra TODO to use `summary.estimatePartitionsNumber `
ex: https://github.com/geotrellis/geotrellis-landsat-tutorial/blob/feature/raster-source/src/main/scala/tutorial/vlm/RasterSourceLocalIngestImage.scala#L60

I'm not sure how big of a deal it is but potentially current method will create very unbalanced partitions.

`sc.parallelize(paths.asScala)` is going to default to `spark.default.parallelism` which I believe defaults for `4` or something like that.

echeipesh: Are you planning to get that in as part of band multiplex feature or do you think it can be an outstanding issue for a bit ?

jbouffard: I was going to include a `numPartitions` parameter in the multiplex PR as a temporary solution. Unless we're going to be using this feature soon, it can probably be an outstanding issue. Though, it's something we should address sooner than later.

echeipesh: Ideally this becomes the way you read all raster in GPS, so it would be worth while to take care of it then.

jbouffard: Agreed. I think it's something we should do next sprint.

PR #694, opened 2018-10-31
Jar Names With Shas

jbouffard:
This PR changes the naming of the jars uploaded to S3 via Travis. Now, the sha of last commit will be included in the jar name unless it's a tagged release. In that case, the version number will be used instead.


jbouffard:
@jpolchlo This should make it  so that the uploaded jars will be public https://github.com/locationtech-labs/geopyspark/blob/752038840d67dd292135f2d8a9d331d65a3a1645/deploy.sh#L6

PR #697, opened 2018-11-29
VectorPipe Removal

jbouffard:
This PR removes the `vector_pipe` submodule and most of its related code. The reason for its removal is because it's becoming harder to support due to its Scala dependencies. In addition, the submodel does not provide useful functionality outside of a few types.

`Feature` and `CellValue` are the only two classes that are being kept. They have been moved into the `geotrellis` submodule.


jbouffard:
I'm closing this PR due to issues encountered with the `geomesa-spark-jts` dependency and the `org.location.jts` vs `com.vividsolutions.jts` conflicts.

geopyspark/tests/vector_pipe/feature_cell_value_schema_test.py
```diff
@@ -1,59 +0,0 @@
...
-from pyspark.serializers import AutoBatchedSerializer
-from geopyspark.vector_pipe import Feature, CellValue
-from geopyspark.geotrellis.protobufserializer import ProtoBufSerializer
-from geopyspark.vector_pipe.vector_pipe_protobufcodecs import (feature_cellvalue_decoder,
-                                                               to_pb_feature_cellvalue,
-                                                               feature_cellvalue_encoder)
-from geopyspark.tests.base_test_class import BaseTestClass
-
-
-class FeatureCellValueSchemaTest(BaseTestClass):
```
echeipesh: These look like they move to `geopyspark/geotrellis/protobufcodecs.py`. Can this test be carried forward?

geopyspark/vector_pipe/protobuf/featureMessages_pb2.py
```diff
@@ -1,352 +0,0 @@
-# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
echeipesh: Are these potentially useful outside of interface with vectorpipe ?

jbouffard: I don't think so. That code is specifically for an older version of `featureMessages.proto`.

geopyspark/tests/vector_pipe/feature_cell_value_schema_test.py
jbouffard: Yeah, it looks like we can still keep that test.

docs/tutorials/reading-and-rasterizing-osm-data.rst
```diff
@@ -1,231 +0,0 @@
...
-   'tiger:zip_right': '19512',
-   'usage': 'freight',
-   'waterway': 'stream'}
-
-
-So it appears that there are schools in this dataset, and that we can continue
-on.
-
-
-Assigning Values to Geometries
```
echeipesh: Looks like rasterization docs that we can keep if we  use non OSM data as example.

PR #698, opened 2018-12-05
Read Multiplex Band Support

jbouffard:
This PR adds multiplex band support to the `TiledRasterLayer.read` method. Now, it'll be possible to read in separate files as a single multiband. This is useful in cases where individual bands are split into separate files. In addition to combing multiple bands into a single tile, it's also possible to specify their order. In order to support band multiplexing, a new type: `SourceInfo` has been created.

*Note:* This PR depends on an unmerged PR: https://github.com/geotrellis/geotrellis-contrib/pull/111 to work.

### DIFF
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
@@ -234,10 +234,10 @@ object RasterSource {
 
     val LayoutLevel(zoom, layout) =
       layoutType match {
-        case global: GlobalLayout =>
+        case global: GPSGlobalLayout =>
           val scheme = ZoomedLayoutScheme(rasterSummary.crs, global.tileSize)
           scheme.levelForZoom(global.zoom)
-        case local: LocalLayout =>
+        case local: GPSLocalLayout =>
           val scheme = FloatingLayoutScheme(local.tileCols, local.tileRows)
           rasterSummary.levelFor(scheme)
       }
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
@@ -48,7 +48,6 @@ object RasterSource {
     rdd: RDD[String],
     targetCRS: String,
     resampleMethod: ResampleMethod,
-    partitionStrategy: PartitionStrategy,
     readMethod: String
   ): ProjectedRasterLayer = {
     val rasterSourceRDD: RDD[RasterSource] =
@@ -84,7 +83,6 @@ object RasterSource {
     paths: java.util.ArrayList[java.util.HashMap[String, String]],
     targetCRS: String,
     resampleMethod: ResampleMethod,
-    partitionStrategy: PartitionStrategy,
     readMethod: String
   ): ProjectedRasterLayer = {
     val scalaPaths: Seq[Seq[(String, String)]] = paths.asScala.toSeq.map { _.asScala.toSeq }
@@ -95,7 +93,6 @@ object RasterSource {
       sc.parallelize(scalaPaths, scalaPaths.size),
       targetCRS,
       resampleMethod,
-      partitionStrategy,
       readMethod
     )
   }
@@ -106,7 +103,6 @@ object RasterSource {
     rdd: RDD[Seq[(String, String)]],
     targetCRS: String,
     resampleMethod: ResampleMethod,
-    partitionStrategy: PartitionStrategy,
     readMethod: String
   ): ProjectedRasterLayer = {
     val rasterSourcesRDD: RDD[Seq[(String, RasterSource)]] =
@@ -167,7 +163,7 @@ object RasterSource {
     sc: SparkContext,
     layerType: String,
     paths: java.util.ArrayList[java.util.HashMap[String, String]],
-    layoutType: LayoutType,
+    layoutType: GPSLayoutType,
     targetCRS: String,
     resampleMethod: ResampleMethod,
     readMethod: String
@@ -181,7 +177,6 @@ object RasterSource {
       layoutType,
       targetCRS,
       resampleMethod,
-      partitionStrategy,
       readMethod
     )
   }
@@ -193,7 +188,6 @@ object RasterSource {
     layoutType: LayoutType,
     targetCRS: String,
     resampleMethod: ResampleMethod,
-    partitionStrategy: PartitionStrategy,
     readMethod: String
   ): SpatialTiledRasterLayer = {
     val rasterSourcesRDD: RDD[Seq[(String, RasterSource)]] =
@@ -269,7 +263,7 @@ object RasterSource {
 
         val groupedTiles: Map[SpatialKey, MultibandTile] =
           groupedValues.map { case (k, vs) =>
-            (k, MultibandTile(vs.map { _._2.band(0) }))
+            (k, MultibandTile(vs.flatMap { _._2.bands }))
           }
 
         groupedTiles.toSeq
--- a/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
+++ b/geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
@@ -87,7 +87,7 @@ object RasterSource {
     partitionStrategy: PartitionStrategy,
     readMethod: String
   ): ProjectedRasterLayer = {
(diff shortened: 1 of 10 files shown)


geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -150,6 +256,131 @@ object RasterSource {
...
+            iter.map { files: Seq[(String, String)] =>
+              files.map { case (k, v) => (k, GDALRasterSource(v): RasterSource)
+              }
+            }
+          }
+      }).cache()
+
+    val orderedRasterSourcesRDD: RDD[Seq[RasterSource]] =
+      rasterSourcesRDD.map { keyedSource: Seq[(String, RasterSource)] =>
+        keyedSource.sortBy { _._1 }.map { case (_, source) => source }
```
echeipesh: I assume by the test that the input is `[(bandOrder, source), ..]`.

For the base use case where you need to ingest 1000 (lets say) Landsat senes you're going to have 1000 sense x 3 bands = 3000 entries in essentially random order. When `sc.parallelize` happens you have no way to know how they were grouped which means any sorting per partition is now partial. You may have gotten entries where all of the bands are "1" or "3" or maybe bands "1", "2", "3" but for scenes that do not spatially overlap.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -150,6 +256,131 @@ object RasterSource {
...
+
+    val tileLayerMetadata: TileLayerMetadata[SpatialKey] =
+      rasterSummary.toTileLayerMetadata(layout, zoom)._1
+
+    val layoutsRDD: RDD[Seq[LayoutTileSource]] =
+      reprojectedSourcesRDD.map { _.map { _.tileToLayout(layout, resampleMethod) } }
+
+    val readInSourcesRDD: RDD[Seq[(SpatialKey, MultibandTile)]] =
+      layoutsRDD.map { layouts: Seq[LayoutTileSource] =>
+        layouts.flatMap { _.readAll() }
```
echeipesh: In reference to above comment: at this point all the tiles are actually read into memory in random order, partially sorted per partition. At some point shuffle needed to happen per `SpatialKey` before this happened.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -150,6 +256,131 @@ object RasterSource {
...
+      targetCRS match {
+        case crs: String =>
+          orderedRasterSourcesRDD.map {
+            _.map { _.reproject(CRS.fromString(crs), resampleMethod) }
+          }
+        case null =>
+          orderedRasterSourcesRDD
+      }
+
+    val rasterSummary: RasterSummary =
```
echeipesh: Now that you have `RasterSummary` is a good chance to use `estimatePartitionsNumber` if no user input to override it was given.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -76,24 +85,121 @@ object RasterSource {
...
+            iter.map { files: Seq[(String, String)] =>
+              files.map { case (k, v) => (k, GDALRasterSource(v): RasterSource)
+              }
+            }
+          }
+      }).cache()
+
+    val orderedRasterSourcesRDD: RDD[Seq[RasterSource]] =
+      rasterSourcesRDD.map { keyedSource: Seq[(String, RasterSource)] =>
+        keyedSource.sortBy { _._1 }.map { case (_, source) => source }
```
echeipesh: sort here is not sufficient. Just due to shape of the input it may be possible to have data for bands "2" and "3" but not "1" for a given location.

echeipesh: There needs to be a hard mapping from source to target bands such that `NODATA` tiles can be generated in above cases.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
jbouffard: :+1:

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
echeipesh: Let's say there are two Landsat scenes, three bands each `(a, b, c)` where all bands for scene 1 and scene 2 are on different partitions. However they both overlap same target tile `SpatialKey(x, y)` in the destination layout. Without a shuffle there are going to be two resulting tiles for `SpatialKey(x, y)`. For things to be in expected format of `TiledRasterSource` they need to be merged I believe.

img_2666

An additional problem is that even restriction each partition to single scene it's easy to read data sizes that are much larger than optimal. All the bands of LC8 are going to be well over a gigabyte in memory when optimal partition sizes appears to be one around 200MB.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -150,6 +270,125 @@ object RasterSource {
...
+
+    val groupedRDD: RDD[(SpatialKey, Iterable[(String, RasterRegion)])] =
+      rasterRegionRDD
+        .groupByKey(SpatialPartitioner(numPartitions.getOrElse(rasterSummary.estimatePartitionsNumber)))
+
+    val tiledRDD: RDD[(SpatialKey, MultibandTile)] =
+      groupedRDD.mapValues { iter =>
+        val sorted = iter.toSeq.sortBy { _._1 }
+
+        MultibandTile(sorted.flatMap { case (_, ref) => ref.raster.toSeq.flatMap { _.tile.bands } })
```
echeipesh: As previously mentioned you can't rely on the fact that a tile for every band is present, therefore `sort` is insufficient. Logic needs to be added to insert `NODATA` tiles for those bands which happen to be missing.

geopyspark-backend/project/Version.scala
```diff
@@ -1,7 +1,7 @@
 object Version {
   val geopyspark = Environment.sha.getOrElse("0.4.2")
   val geotrellis = "3.0.0-SNAPSHOT"
```
echeipesh: Should use `2.2.0` now that it is released, there is corresponding version of geotrellis-contrib-vlm that is available.

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -29,17 +30,25 @@ object RasterSource {
...
+      numPartitions match {
+        case i: Integer => Some(i.toInt)
+        case null => None
+      }
+
     read(
       sc,
       layerType,
-      sc.parallelize(paths.asScala),
+      sc.parallelize(paths.asScala, partitions.getOrElse(sc.defaultParallelism)),
```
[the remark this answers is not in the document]
jbouffard: :+1:

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -76,24 +85,135 @@ object RasterSource {
...
+            }
+          }
+        case GDAL =>
+          rdd.mapPartitions { iter =>
+            iter.map { files: Seq[(String, String)] =>
+              files.map { case (k, v) => (k, GDALRasterSource(v): RasterSource)
+              }
+            }
+          }
+      }).cache()
```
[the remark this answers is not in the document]
jbouffard: :+1:

geopyspark-backend/geotrellis/src/main/scala/geopyspark/geotrellis/vlm/RasterSource.scala
```diff
@@ -76,24 +85,135 @@ object RasterSource {
...
+
+        val tiles: Seq[MultibandTile] =
+          sorted.flatMap { case (_, source) =>
+            source.read(source.extent) match {
+              case Some(raster) => Some(raster.tile)
+              case None => None
+            }
+          }
+
+        (ProjectedExtent(ex, crs), MultibandTile(tiles.map { _.band(0) }))
```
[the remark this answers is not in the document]
jbouffard: Ah, I was assuming that that the inputs would be single band tiles. Though, I'm now realizing that might not always be the case.

